Issue #3, opened 2019-03-28
Database-Add-Remove


Daniel-Jang:
We want to make sure that our Django app can retrieve, add and remove records from the connected PostgreSQL server. This is less about implementing actual functionality, but it is just to learn how to deal with the database.

Acceptance Criteria:
- There exists some backend code that can retrieve, add and remove records to and from a table created in PostgreSQL.

Daniel-Jang:
Met the acceptance criteria; created pull request here: https://github.com/cass-degrees/CASS-Degrees-Code/pull/4

PR #4, opened 2019-04-02
Add, edit and remove records from a table in a postgresql database

Daniel-Jang:
We now can add, edit and remove records from a sample model I created (primary key 'id' and a text field 'text') using the APIs from the Django REST Framework.

Relating issue: https://github.com/cass-degrees/CASS-Degrees-Code/issues/3

### DIFF
--- a/cassdegrees/api/admin.py
+++ b/cassdegrees/api/admin.py
@@ -1,3 +1,5 @@
 from django.contrib import admin
+from .models import SampleModel
 
 # Register your models here.
+admin.site.register(SampleModel)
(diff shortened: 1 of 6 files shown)


Daniel-Jang:
Perform database migrations before testing:

python manage.py makemigrations
python manage.py migrate

JackSekoranja (approved):
Works nicely. I tested Adding/editing/viewing a new record, as well as viewing/editing/deleting on an admin account and found no issues :)

aidengeorge97 (approved):
Database is set up appropriately and can add/remove records as the feature requires.

Issue #38, opened 2019-04-07
Define-Template-Structure


Daniel-Jang:
Structure the templates folder so it is clear where to place the CSS, javascript and image files.

Acceptance Criteria:
- The ~~templates~~ newly created 'static' folder should be well structured so that it is clear where to place different kinds of files (CSS, js, image files etc).
- settings.py must be modified to recognise the files in the static folder.

Daniel-Jang:
Ok so I've met the criteria, you can review it by checking index.html and the static folder (and settings.py if you want to) and run Django to confirm it works (right click and inspect to see if style.css is loaded).

Also, there are more resources regarding this:
https://docs.djangoproject.com/en/2.1/intro/tutorial06/
https://docs.djangoproject.com/en/2.1/howto/static-files/

Daniel-Jang:
I've submitted a pull request here: https://github.com/cass-degrees/CASS-Degrees-Code/pull/46

PR #46, opened 2019-04-08
Create and register a folder that contains the CSS, img and js files

Daniel-Jang:
Refer to the original issue here for more information: https://github.com/cass-degrees/CASS-Degrees-Code/issues/38


JackSekoranja (approved):
Seems to work fine and html/css locations are mostly clear. It did throw an exception initially when I navigated to `/api/sample/`, but this resolved itself when I reran `python manage.py makemigrations` and `python manage.py migrate`

@Daniel-Jang Note that Git doesn't keep track of directories, so the _img_ and _js_ directories weren't included in the commit you added

Daniel-Jang:
As Jack mentioned, turns out git doesn't add empty directories, so I added some placeholder files. When actual images and javascript files are added, please delete these placeholder files.

JackSekoranja:
Deleting the files will update Git and delete the folders, so you'll have to do this AFTER adding a different file in the directory

j-selby (approved):
The conventional way to keep an otherwise empty directory is a `.gitkeep` file, but other than that, looks good!

Issue #45, opened 2019-04-08
Work on Styling Templates


JackSekoranja:
• Work on styling templates for other frontend devs to use
• Generate Django templates based on styling (allowing for consistency)

j-selby:
Base template implemented as of #47.

PR #47, opened 2019-04-09
Implement base template conformant to ANU styling

j-selby:
This commit adds a base template which pulls in ANU styling, and provides a basic, configurable layout which can be utilised by pages.

Partially completes #45

Example:

image

### DIFF
--- a/cassdegrees/static/css/style.css
+++ b/cassdegrees/static/css/style.css
@@ -1,5 +1,5 @@
 @charset "UTF-8";
 
-body {
-    border: 1px;
+#body {
+    min-height: 600px;
 }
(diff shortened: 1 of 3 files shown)


Daniel-Jang:
I had a look and I don't see any problems with what you have added. It seems to be very similar to the template example Django provided which is good: 
https://docs.djangoproject.com/en/2.1/ref/templates/language/

While it would be ideal for 2 reviewers' opinions, I personally think you successfully implemented base and index.

Daniel-Jang (approved):

j-selby:
@Daniel-Jang Correct - no significant templating info out there for Django, so had to use that as a guide. Let me note that.

JackSekoranja:
I cant review this reques but I've checked this and it seems to work well. I've used it as part of my branch and I've used it with no issues.

JackSekoranja (approved):
Changes seem to work well. I've used it as part of my branch without any issues.

Issue #39, opened 2019-04-07
UI-API-Interaction-Demo


Daniel-Jang:
Create a simple web page with fields and buttons that once a 'submit' button is pressed, the written information gets saved to the 'sample' model. This is in direct preparation for the course editing and removal user stories.

Acceptance Criteria:
- Once a 'submit' button is hit, the UI hits the API endpoint (as implemented in https://github.com/cass-degrees/CASS-Degrees-Code/issues/3) with appropriate HTML requests.
- Relevant information filled in the web page is added to the database.

Issue #31, opened 2019-04-03
Course Editing (5)


JackSekoranja:
As a CASS Staff member,
I want to be able to edit courses that are offered in the college
So that I can correct any mistakes that were made when creating the course.

**Acceptance Criteria:**
- The user is able to select a course and modify the name, the requisites, the year it runs in, and the semesters it runs in
- The changes are updated in the system after saving

Daniel-Jang:
Just a note, this is currently blocked until this pull request (https://github.com/cass-degrees/CASS-Degrees-Code/pull/58) is accepted and merged back into develop. I need to edit code in managecourses.html and manage_courses function in ui/views.py.

I will be helping out Spencer with that PR and then I will be able to work on this.

PR #48, opened 2019-04-10
Add, edit and delete records from the sample model using a custom-made very basic UI Form

Daniel-Jang:
This pull request directly relates to: https://github.com/cass-degrees/CASS-Degrees-Code/issues/39
and is preparing for: https://github.com/cass-degrees/CASS-Degrees-Code/issues/31 once the schemas are ready.

To check it out, navigate to 127.0.0.1:8000/sampleform/

### DIFF
--- a/cassdegrees/templates/sampleform.html
+++ b/cassdegrees/templates/sampleform.html
@@ -5,7 +5,7 @@
 {% block subtitle %}Add, edit or delete sample records{% endblock %}
 
 {% block content %}
-    <form class="anuform labelwide" action="/sampleform/" method="post">
+    <form class="anuform" action="/sampleform/" method="post">
         {% csrf_token %}
         <fieldset>
             <legend>Add Record</legend>
@@ -24,12 +24,12 @@
                 <input class="text" id="text" type="text" name="text" aria-required="true" required>
             </p>
         </fieldset>
-        <p>
+        <p class="text-right">
             <input class="btn-uni-grad btn-large" type ="submit" value="Create">
         </p>
     </form>
     <br />
-    <form class="anuform labelwide" action="/sampleform/" method="post">
+    <form class="anuform" action="/sampleform/" method="post">
         {% csrf_token %}
         <fieldset>
             <legend>Modify Record</legend>
@@ -48,13 +48,13 @@
                 <input class="text" id="text" type="text" name="text" aria-required="true" required>
             </p>
         </fieldset>
-        <p>
+        <p class="text-right">
             <input class="btn-uni-grad btn-large" type ="submit" value="Update">
         </p>
 
     </form>
     <br />
-    <form class="anuform labelwide" action="/sampleform/" method="post">
+    <form class="anuform" action="/sampleform/" method="post">
         {% csrf_token %}
         <fieldset>
             <legend>Delete Record</legend>
@@ -68,7 +68,7 @@
                 <input class="text" id="id" type="text" name="id" aria-required="true" required>
             </p>
         </fieldset>
-        <p>
+        <p class="text-right">
             <input class="btn-uni-grad btn-large" type ="submit" value="Delete">
         </p>
     </form>
--- a/cassdegrees/templates/sampleform.html
+++ b/cassdegrees/templates/sampleform.html
@@ -5,12 +5,46 @@
     <title>Title</title>
 </head>
 <body>
+--------------- Add record ----------------
     <form action="/sampleform/" method="post">
         {% csrf_token %}
+        <!-- As HTML forms only support get or post methods,
+        add this hidden input for consistency with the rest of the forms -->
+        <input type="hidden" name="_method" value="post">
+
         <label for="id">ID:</label>
         <input id="id" type="text" name="id">
+
         <label for="text">Text:</label>
         <input id="text" type="text" name="text">
+
+        <input type ="submit" value="Submit">
+    </form>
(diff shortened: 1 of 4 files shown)


JackSekoranja (changes requested):
Since you are going to use a similar format for the user stories, I think we should use the correct way now. Refer to the following page as a reference for formatting the forms correctly: http://webpublishing.anu.edu.au/web-style-guide/forms.php

Daniel-Jang:
Ok, followed Jack's suggestion and now the page looks like this:
image

cassdegrees/templates/sampleform.html
```diff
@@ -0,0 +1,75 @@
...
+            <p>
+                <label for="id">ID:</label>
+                <input class="text" id="id" type="text" name="id" aria-required="true" required>
+            </p>
+        </fieldset>
+        <p>
+            <input class="btn-uni-grad btn-large" type ="submit" value="Delete">
+        </p>
+    </form>
+{% endblock %}
```
j-selby: Newline at end of file

cassdegrees/templates/sampleform.html
```diff
@@ -0,0 +1,75 @@
...
+            add this hidden input to hint that it is actually a delete request. -->
+            <input type="hidden" name="_method" value="delete">
+
+            <p>
+                <label for="id">ID:</label>
+                <input class="text" id="id" type="text" name="id" aria-required="true" required>
+            </p>
+        </fieldset>
+        <p>
+            <input class="btn-uni-grad btn-large" type ="submit" value="Delete">
```
j-selby, line 72: Buttons should be right-aligned as per styling guide. (x3 for this + above)

cassdegrees/ui/views.py
```diff
@@ -1,8 +1,66 @@
...
     return render(request, 'index.html', context={})
+
+
+# I went through this tutorial to create the form html file and this view:
+# https://docs.djangoproject.com/en/2.2/topics/forms/
+# Hope this serves as an inspiration for when we make proper views and functions to submit course information
+def sampleform(request):
+    # If POST request, redirect the received information to the backend:
+    if request.method == 'POST':
+        model_api_url = 'http://127.0.0.1:8000/api/sample/'
```
j-selby, line 18: Hmmm.... Not sure I like the idea of bouncing stuff around via HTTP internally - no way to directly talk to the API?

j-selby (changes requested):
Looking good!

JackSekoranja (approved):

cassdegrees/ui/views.py
Daniel-Jang: Not that I know of, the API from the Django REST Framework is built to accept incoming POST, PUT, PATCH and DELETE requests so I used them.

Daniel-Jang:
Made changes suggested by James, except the API interaction by URL (commented on that suggestion)

New look:
image

cassdegrees/ui/views.py
j-selby: The problem is overhead + port changes + etc here. We are effectively handling double HTTP requests here now.

Daniel-Jang: Keep in mind that our UI and API are considered to be separate apps, so importing the API related views to the UI application doesn't make much sense. Since they are separate apps, it makes sense to submit HTTP requests to the API endpoint.

My idea here was to receive the data from the user input from the form, package it nicely in the UI app to a format the API (api/sample/) will accept and then submit a HTTP request to the API endpoint.

Have a look at the link below:
https://stackoverflow.com/questions/11663945/calling-a-rest-api-from-django-view

Regarding your point on overhead, could you please clarify what kind of overhead (and where) you are talking about?

Hard coding URL is a bad idea that I do admit, but we have no better method of dealing with that as of this moment and is not relevant to this pull request.

Double HTTP requests are of little concern (minimal performance impact), the data must be transferred to the API regardless of method (whether function calls or HTTP request).

However, I am open to suggestions, please feel free to let me know if you find a better method of going about this.

j-selby: In terms of overhead, for each of these API requests:
- A new TCP connection has to be created, including operating system overhead (lookup, routing, etc), which is a non-zero cost for network connections.
- HTTP request + response parsing
- Domain filtering (i.e ALLOWED_HOSTS)

If we cannot talk between the API and UI directly, would unix sockets also be a solution?

Daniel-Jang: I am not familiar at all with Unix sockets, but I'm not sure if it would be a solution here as "A UNIX socket is an inter-process communication mechanism that allows bidirectional data exchange between processes running on the same machine." - https://serverfault.com/questions/124517/whats-the-difference-between-unix-socket-and-tcp-ip-socket.

UI and API are not actual standalone processes but are simply separate web-apps. While I said you can't import things from API to UI, you can but it is not good practice to be doing that and I am not aware of any Python functions of ListAPIView (api/sample/) that you can call to insert data.

But if we choose to go down this route, this means that we will have to create listeners, learn to understand Unix socket data transmissions etc for some gain that in reality, no one will notice. In my opinion, any benefit gained by going this route will be far outweighed by the things I have said in the previous sentence.

j-selby: That is fair then. Fairly disappointing, but hopefully this doesn't become an issue.

j-selby (approved):

Issue #41, opened 2019-04-07
Create Empty Data Tables


JackSekoranja:
This task will be to implement the schema of the database so that we can start filling them in.

aidengeorge97:
Created tables for storing static Course, Degree, and Subplan information. Data can now be stored in these tables, and courses can be linked to subplans. 

Not included:
- Tight restrictions on variable values. 
- Functionality to store degree rules
- Functionality to store requisites and incompatibilities fro courses

PR #49, opened 2019-04-10
Create Empty Data Tables

aidengeorge97:
Refer to the original issue here for more information #41. 

Added functionality for storing basic data sets in the database. Does not contain functionality for storing rules or course requisites/incompatibilities.


JackSekoranja (approved):
Works nicely. I should note that the representation of courses (specifically including the year and semsters offered) may be strictly make sense in the context we will be using it for. For example, if we say a major requires the completion of a course, any offering of that course would be valid, not one in an arbitrary semester. Despite this, the new tables should be sufficient for all of our uses.

Daniel-Jang:
Functionality wise, it seems good to me but is there a reason why the max length for course codes is 32 characters long? The same goes for year and units, which can be set ridiculously high:
image

All of these potential extreme values can be remedied by Python code (e.g. if len(course_code) == 8) but do you think it would be better to leave the models as is? Or would restricting possible values be better (course code max_length = 8 for example).

Also, I think that the urls for these end points should be something like /api/model/course/ instead of /api/course/ for example. We will eventually be creating custom API views that isn't focused on database transactions and having all urls at the root level may cause confusion later. I know I originally didn't add 'model' either so that is my bad.

aidengeorge97:
I can shrink down the max sizes, but it felt safe to leave a little extra room for edge cases and we're not short on storage space anyway. 

I agree on the URL links. I can make that change now.

joshcbee (approved):
Basic functionality is working well to get started on other dependencies. Only point to question is whether we want a drop-down selector for Degree Type or free-text (noting from the user story/meeting discussion they'd like the flexibility to specify type e.g. Adding a Diploma

Following on from Daniel's point around constraints, it may be worth integrating the following in the next iteration:
- year field constraint for four digits
- all units must be greater than zero (possible to do divisible by 6 as a database constraint?)
- subplan codes should follow pattern of P&C, i.e. either "XXXX-MAJ, XXXX-MIN, XXXX-SPEC' or just 'XXXX' if we can already generate the suffix from the sub-plan type. This would help us generate P&C links later.
- course codes should be 8 long, if possible 4 char, 4 integers

aidengeorge97:
Good feedback. I will merge this branch so others can make use of it and consider those suggestions for future development.

Issue #40, opened 2019-04-07
Implement a PEP 8 Checker


Daniel-Jang:
Implement some kind of a pep 8 ([Python code style guide](https://www.python.org/dev/peps/pep-0008/)) checker that checks the entire code base and reports any pep 8 violations.

Acceptance Criteria:
- Implement a pep 8 checker (as per description).
- Customisable. Allows exceptions and modifications to the default rules in the pep 8 checker of choice.

j-selby:
Implemented in #54

PR #54, opened 2019-04-11
Implemented flake8 and desired settings in setup.cfg

Daniel-Jang:
To check the entire codebase for pep 8 violations, simply type:

flake8 

in the console. If nothing shows up, that means there are no violations.

Relevant issue: https://github.com/cass-degrees/CASS-Degrees-Code/issues/40


cassdegrees/setup.cfg
```diff
@@ -0,0 +1,8 @@
+[flake8]
+# F401: Ignore unused imports - some files are automatically generated with import statements.
+# F403, F405: Ignore warnings caused by * imports
+ignore = F401, F403, F405
+# Ignore the git files and migration files that are out of our control.
+exclude = .git, *migrations*
+# Set max length to be the same as what PyCharmm currently has it at (120).
+max-line-length = 120
```
j-selby: Newline.

j-selby:
Hmmm. Would this be better serviced with a GitHub app (e.g. https://github.com/marketplace/actions/python-lint ) so we don't have to manually do this every time?

j-selby:
Specifically https://github.com/marketplace/actions/python-syntax-checker might be what we are looking for here.

Daniel-Jang:
This allows us to check for code violations and fix all of them before commit. 

If we go your method only, we'd have to commit our code, look at status of CI, come back to edit all the changes, commit again, check if further issues popped up and repeat until no pep 8 errors remain.

What I suggest is that we have both this and the CI. How does that sound?

j-selby:
Nope, that sounds perfectly fine! Just tested it out as above, but it will need a bit of configuration

Daniel-Jang:
Did you mean the local flake8? If yes, here it is: https://github.com/cass-degrees/CASS-Degrees-Code/pull/54/commits/42d1d165a2dd34dcd422f00cafe6be78f5e0c628#diff-9a1f66670f0a343dc88bff1ada2c310d
I added comments on the reasoning as well.

If you meant the CI pep 8 checker, I will try and see how I can add configurations to it. I'll be adding the exact configurations as I had it with the local flake8.
(note to self: https://pep8speaks.com/)

j-selby:
Yeah. I meant the CI - should just be an additional configuration file to merge here.

JackSekoranja (approved):
Ran it and it works, although note that it should be run in the `CASS-Degrees-Code/cassdegrees` directory

Daniel-Jang:
Ok, I added the .pep8speaks.yml configuration file with some settings.

.pep8speaks.yml
```diff
@@ -0,0 +1,10 @@
+scanner:
+  diff_only: True
+  linter: flake8
+
+flake8:
+  max-line-length: 120  # Set max length to be the same as what PyCharmm currently has it at (120).
+  ignore:
+    - F401  # Ignore unused imports - some files are automatically generated with import statements.
+    - F403  # F403, F405: Ignore warnings caused by * imports
+    - F405
```
j-selby, line 10: Newline :P

j-selby (approved):
Looks good!

.pep8speaks.yml
Daniel-Jang: Oh man, I noticed this comment after merging it...

Haha damn sorry again.... I'll fix it in the next branch I work on :/

PR #57, opened 2019-04-11
Data list page improvements

JackSekoranja:
Made modifications to my previously created list page to make it more dynamic and include "Courses" as one of the tabs. The Idea here is we will be able to use it as the endpoint for most of our searches for existing content (As such, I have also linked the page to the "Manage Courses" section). I've also made it possible to open a default tab on this page by specifying `?view=OPT` (OPT = Degree, Subplan, Course).

New List Page

### DIFF
--- a/cassdegrees/templates/list.html
+++ b/cassdegrees/templates/list.html
@@ -1,8 +1,8 @@
 {% extends "base.html" %}
 
-{% block page-title %}Template/Subplan List{% endblock %}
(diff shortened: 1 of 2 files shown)


j-selby (approved):
Looks good so far, though obviously individual types do need to be parsed (e.g. booleans).

Daniel-Jang:
Nice work!

Daniel-Jang (approved):

Issue #27, opened 2019-04-03
Course Creation and Specification (5)


JackSekoranja:
As a CASS Staff member, 
I want to be able to add new courses to the system,
So that they can be used in future program plans or subplans

**Acceptance Criteria:**
- The user is able to add a new course to the system by specifying the name, the course code, the year it runs in, and the semesters it runs in
- The user is able to specify the requisite/incompatible courses for the entered course
- If the requisite/incompatible courses aren’t in the system after the user saves the new course, the user is prompted to fill in the course details for the requisite/incompatible course
- The user is notified if the course they are adding already exists
- The course is present in the system after saving

PR #58, opened 2019-04-12
Course Creation and Specification

spencerj411:
Completed Issue #27 (Course Creation and Specification (5)).

Can access this page at `http://127.0.0.1:8000/manage_courses/`. It displays a form in 3 sections (for this user story, only 'Add Course' has been implemented).

**The acceptance criterion:** 
> If the requisite/incompatible courses aren’t in the system after the user saves the new course, the user is prompted to fill in the course details for the requisite/incompatible course

has been satisfied by listing all of the courses in the database instead of giving the option to the user to type the courses manually.
> Notify the user if they are attempting to add an existing course

was implemented with a basic HttpResponse.

**Small things:**
1. I was not sure how to concatenate 3 strings together hence `{{ course.code }}_{{ course.year }}` appeared 4 times in `managecourses.html` (solved)
2. Creation of courses can be tested by checking `http://127.0.0.1:8000/list/?view=Course`
3. Screenshot of UI  (outdated, check below for a more updated screenshot)
image

### DIFF
--- a/cassdegrees/static/css/style.css
+++ b/cassdegrees/static/css/style.css
@@ -27,3 +27,11 @@
     justify-content: center;
     margin: 12px;
 }
+
+select.tfull {
+    width: 70%;
+}
+
+select.text{
+    border-radius: 3px;
+}
--- a/cassdegrees/static/css/style.css
+++ b/cassdegrees/static/css/style.css
@@ -29,7 +29,7 @@
 }
 
 select.tfull {
-    width: 70%;
+    width: 70.25%;
 }
 
 select.text{
--- a/cassdegrees/templates/managecourses.html
+++ b/cassdegrees/templates/managecourses.html
@@ -31,7 +31,7 @@
             </p>
 
             <p>
-                <label for="text">Semester:</label>
+                <label for="text">Semester(s):</label>
                 <select name="semester[]" size="2" multiple>
                     <option value="semester1">Semester 1</option>
                     <option value="semester2">Semester 2</option>
@@ -41,6 +41,24 @@
              <p>
                 <label for="text">Units:</label>
                 <input class="text" id="number" type="number" name="units" min="6" max="48" step="6" aria-required="true" required>
+            </p>
+
+            <p>
+                <label for="text">Requisite Course(s):</label>
+                <select name="semester[]" size="5" multiple>
+                    {% for course in courses %}
+                        <option value="{{ course.code }}_{{ course.year }}">{{ course.code }}_{{ course.year }}</option>
+                    {% endfor %}
+                </select>
+            </p>
+
+            <p>
+                <label for="text">Incompatible Course(s):</label>
+                <select name="semester[]" size="5" multiple>
+                    {% for course in courses %}
+                        <option value="{{ course.code }}_{{ course.year }}">{{ course.code }}_{{ course.year }}</option>
+                    {% endfor %}
+                </select>
             </p>
         </fieldset>
         <p class="text-right">
--- a/cassdegrees/templates/managecourses.html
+++ b/cassdegrees/templates/managecourses.html
@@ -57,7 +57,7 @@
                 <label for="text">Incompatible Course(s):</label>
                 <select name="semester[]" size="5" multiple>
                     {% for course in courses %}
-                        <option value="{{ course.code }}_{{ course.year }}">{{ course.code }}_{{ course.year }}</option>
+                        <option value="{{ course.code }}">{{ course.code }}</option>
                     {% endfor %}
                 </select>
             </p>
--- a/cassdegrees/templates/managecourses.html
+++ b/cassdegrees/templates/managecourses.html
@@ -15,22 +15,22 @@
             <input type="hidden" name="_method" value="post">
 
             <p>
-                <label class="req" for="id">Course Name:</label>
+                <label class="req" for="id">Course Name</label>
                 <input class="text tfull" id="text" type="text" name="name" aria-required="true" required>
             </p>
 
             <p>
-                <label class="req" for="id">Course Code:</label>
+                <label class="req" for="id">Course Code</label>
                 <input class="text tfull" id="text" type="text" name="code" minlength="8" maxlength="8" size="8" aria-required="true" required>
             </p>
 
             <p>
-                <label class="req" for="text">Year:</label>
+                <label class="req" for="text">Year</label>
                 <input class="text tfull" id="number" type="number" name="year" min="1990" aria-required="true" required>
             </p>
 
             <p>
-                <label for="text">Semester(s):</label>
+                <label for="text">Semester(s)</label>
                 <select class="text tfull" name="semester[]" size="2" multiple>
                     <option value="semester1">Semester 1</option>
                     <option value="semester2">Semester 2</option>
@@ -38,13 +38,13 @@
             </p>
 
              <p>
-                <label class="req" for="text">Units:</label>
+                <label class="req" for="text">Units</label>
                  <!-- max value is chosen arbitrarily -->
                 <input class="text tfull" id="number" type="number" name="units" min="0" step="6" aria-required="true" required>
             </p>
 
             <p>
-                <label for="text">Requisite Course(s):</label>
+                <label for="text">Requisite Course(s)</label>
                 <select class="text tfull" name="semester[]" size="5" multiple>
                     {% for course in courses %}
                         <option value="{{ course.code }}">{{ course.code }}</option>
@@ -53,7 +53,7 @@
             </p>
 
             <p>
-                <label for="text">Incompatible Course(s):</label>
+                <label for="text">Incompatible Course(s)</label>
                 <select class="text tfull" name="semester[]" size="5" multiple>
                     {% for course in courses %}
                         <option value="{{ course.code }}">{{ course.code }}</option>
--- a/cassdegrees/templates/managecourses.html
+++ b/cassdegrees/templates/managecourses.html
@@ -41,7 +41,7 @@
              <p>
                 <label for="text">Units:</label>
                  <!-- max value is chosen arbitrarily -->
-                <input class="text" id="number" type="number" name="units" min="6" max="48" step="6" aria-required="true" required>
+                <input class="text tfull" id="number" type="number" name="units" min="0" step="6" aria-required="true" required>
             </p>
 
             <p>
--- a/cassdegrees/templates/managecourses.html
+++ b/cassdegrees/templates/managecourses.html
@@ -21,7 +21,7 @@
 
             <p>
                 <label for="id">Course Code:</label>
-                <input class="text" id="text" type="text" name="code" minlength="8" maxlength="8" size="8" aria-required="true" required>
+                <input class="text tfull" id="text" type="text" name="code" minlength="8" maxlength="8" size="8" aria-required="true" required>
             </p>
 
             <p>
--- a/cassdegrees/templates/managecourses.html
+++ b/cassdegrees/templates/managecourses.html
@@ -8,7 +8,7 @@
     <form class="anuform" action="/manage_courses/" method="post">
         {% csrf_token %}
         <fieldset>
-            <legend>Add Courses</legend>
+            <legend>Add Course</legend>
 
             <!-- As HTML forms only support get or post methods,
             add this hidden input for consistency with the rest of the forms -->
--- a/cassdegrees/templates/managecourses.html
+++ b/cassdegrees/templates/managecourses.html
@@ -15,17 +15,17 @@
             <input type="hidden" name="_method" value="post">
 
             <p>
-                <label for="id">Course Name:</label>
+                <label class="req" for="id">Course Name:</label>
                 <input class="text tfull" id="text" type="text" name="name" aria-required="true" required>
             </p>
 
             <p>
-                <label for="id">Course Code:</label>
+                <label class="req" for="id">Course Code:</label>
                 <input class="text tfull" id="text" type="text" name="code" minlength="8" maxlength="8" size="8" aria-required="true" required>
             </p>
 
             <p>
-                <label for="text">Year:</label>
+                <label class="req" for="text">Year:</label>
                 <input class="text tfull" id="number" type="number" name="year" min="1990" aria-required="true" required>
             </p>
 
@@ -38,7 +38,7 @@
             </p>
 
              <p>
-                <label for="text">Units:</label>
+                <label class="req" for="text">Units:</label>
                  <!-- max value is chosen arbitrarily -->
                 <input class="text tfull" id="number" type="number" name="units" min="0" step="6" aria-required="true" required>
             </p>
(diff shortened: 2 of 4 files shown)


JackSekoranja (changes requested):
Form looks really good, but since adding and editing should be the same form we may be able to simplify it using a similar trick to my `list` page, so I think leave the additional forms out of it for now. There were also a few minor changes that I will suggest below.

cassdegrees/ui/views.py
```diff
@@ -95,3 +95,66 @@ def sampleform(request):
...
     return render(request, 'createsubplan.html')
+
+
+# inspired by the samepleform function creatd by Daniel Jang
+def manage_courses(request):
+    courses = requests.get(request.build_absolute_uri('/api/model/course/?format=json')).json()
+    # If POST request, redirect the received information to the backend:
+    if request.method == 'POST':
+        # hard coded url; only temporary
+        model_api_url = 'http://127.0.0.1:8000/api/model/course/'
```
JackSekoranja:
```suggestion
        model_api_url = request.build_absolute_uri('/api/model/course/')
```

This Makes it a relative path, meaning we won't have to change it later (We should probably do this in `sampleform` above too)

cassdegrees/templates/managecourses.html
```diff
@@ -0,0 +1,115 @@
...
+
+            <p>
+                <label for="id">Course Code:</label>
+                <input class="text" id="text" type="text" name="code" minlength="8" maxlength="8" size="8" aria-required="true" required>
+            </p>
+
+            <p>
+                <label for="text">Year:</label>
+                <!-- static min max values are only temporary -->
+                <input class="text" id="number" type="number" name="year" min="1990" max="2100" aria-required="true" required>
```
JackSekoranja:
```suggestion
                <input class="text tfull" id="number" type="number" name="year" min="1990" aria-required="true" required>
```

Changes will make the form stretch across to the end of the page. Also remove max year, it isn't needed.

cassdegrees/templates/managecourses.html
```diff
@@ -0,0 +1,115 @@
...
+        <fieldset>
+            <legend>Add Course</legend>
+
+            <!-- As HTML forms only support get or post methods,
+            add this hidden input for consistency with the rest of the forms -->
+            <input type="hidden" name="_method" value="post">
+
+            <p>
+                <label for="id">Course Name:</label>
+                <input class="text" id="text" type="text" name="name" aria-required="true" required>
```
JackSekoranja:
```suggestion
                <input class="text tfull" id="text" type="text" name="name" aria-required="true" required>
```

Makes the text box stretch to the end of the page, giving the user more space to type

cassdegrees/templates/managecourses.html
```diff
@@ -0,0 +1,115 @@
...
+            </p>
+
+            <p>
+                <label for="id">Course Code:</label>
+                <input class="text" id="text" type="text" name="code" minlength="8" maxlength="8" size="8" aria-required="true" required>
```
JackSekoranja:
```suggestion
                <input class="text tfull" id="text" type="text" name="code" minlength="8" maxlength="8" size="8" aria-required="true" required>
```

Same as above (for consistency)

cassdegrees/templates/managecourses.html
```diff
@@ -0,0 +1,115 @@
...
+
+            <p>
+                <label for="text">Year:</label>
+                <!-- static min max values are only temporary -->
+                <input class="text" id="number" type="number" name="year" min="1990" max="2100" aria-required="true" required>
+            </p>
+
+            <p>
+                <label for="text">Semester(s):</label>
+                <select name="semester[]" size="2" multiple>
```
JackSekoranja: I can't work out how to make this stretch to the end of the screen, but it would be good to work out

cassdegrees/templates/managecourses.html
```diff
@@ -0,0 +1,115 @@
...
+                    <option value="semester1">Semester 1</option>
+                    <option value="semester2">Semester 2</option>
+                </select>
+            </p>
+
+             <p>
+                <label for="text">Units:</label>
+                 <!-- max value is chosen arbitrarily -->
+                <input class="text" id="number" type="number" name="units" min="6" max="48" step="6" aria-required="true" required>
```
JackSekoranja:
```suggestion
                <input class="text tfull" id="number" type="number" name="units" min="0" step="6" aria-required="true" required>
```

Courses can be worth 0 units, so this should be reflected here.
Stretch to end of form and removed max units for a course (Realistically it should only be 12-24 max but why restrict them?)

cassdegrees/templates/managecourses.html
```diff
@@ -0,0 +1,115 @@
...
+            </p>
+
+            <p>
+                <label for="text">Requisite Course(s):</label>
+                <select name="semester[]" size="5" multiple>
+                    {% for course in courses %}
+                        <option value="{{ course.code }}_{{ course.year }}">{{ course.code }}_{{ course.year }}</option>
```
JackSekoranja:
```suggestion
                        <option value="{{ course.code }}">{{ course.code }}</option>
```

Course year is not relevant to requisites, so I have removed them (Need to ensure this is also done when passing the API to prevent repeat values). This also raises another important issue, which is that the requisites is not currently represented in the database. This may be a problem for @aidengeorge97 to solve

cassdegrees/templates/managecourses.html
```diff
@@ -0,0 +1,115 @@
...
+                    {% endfor %}
+                </select>
+            </p>
+
+            <p>
+                <label for="text">Incompatible Course(s):</label>
+                <select name="semester[]" size="5" multiple>
+                    {% for course in courses %}
+                        <option value="{{ course.code }}_{{ course.year }}">{{ course.code }}_{{ course.year }}</option>
```
JackSekoranja:
```suggestion
                        <option value="{{ course.code }}">{{ course.code }}</option>
```

As above

cassdegrees/templates/managecourses.html
```diff
@@ -0,0 +1,115 @@
...
+                    {% endfor %}
+                </select>
+            </p>
+        </fieldset>
+        <p class="text-right">
+            <input class="btn-uni-grad btn-large" type ="submit" value="Create">
+        </p>
+    </form>
+    <br />
+    <!-- not yet implemented -->
```
JackSekoranja: Might be a good idea to remove everything after this line until the `{% endblock %}` line. `Delete` could have it's own form, but the `edit` form would be the exact same as the one above, only with pre-filled values and with the `value=patch` item. Ask me if unsure about how to approach for future stories

cassdegrees/templates/managecourses.html
```diff
@@ -0,0 +1,115 @@
...
+
+{% block subtitle %}Add, Modify or Delete Courses{% endblock %}
+
+{% block content %}
+    <form class="anuform" action="/manage_courses/" method="post">
+        {% csrf_token %}
+        <fieldset>
+            <legend>Add Course</legend>
+
+            <!-- As HTML forms only support get or post methods,
```
JackSekoranja: General note about comments in this page. When you comment with `<!-- ... -->` it appears when you view the page source, but when you use Django Template Comments (`{# ... #}`) they get deleted when the page is hosted.

cassdegrees/ui/views.py
```diff
@@ -95,3 +95,66 @@ def sampleform(request):
 
 def create_subplan(request):
     return render(request, 'createsubplan.html')
+
+
+# inspired by the samepleform function creatd by Daniel Jang
+def manage_courses(request):
+    courses = requests.get(request.build_absolute_uri('/api/model/course/?format=json')).json()
```
JackSekoranja, line 108:
```suggestion
    courses = requests.get(request.build_absolute_uri('/api/model/course/?format=json')).json()
    courses = [{'code': course} for course in  set([x['code'] for x in courses])]
```

Convert courses to a unique set of only course codes. Please note that this is NOT a good solution and is intended to be a single line fix. Best way might be to try do a SQL Query for unique course codes, consult @Daniel-Jang for this

JackSekoranja (changes requested):
These are my suggested changes

cassdegrees/templates/managecourses.html
spencerj411: my thought process was so it leaves the user less of a chance to create a mistake when creating a course, hence adding max for both year and units.

cassdegrees/templates/managecourses.html
spencerj411: my thought process was so it leaves the user less of a chance to create a mistake when creating a course, hence adding max for both year. My original intention was to make it dynamic but will do that at a later date once the main work has been approved and if everyone is ok with this.

cassdegrees/templates/managecourses.html
spencerj411: good idea

cassdegrees/templates/managecourses.html
spencerj411: not particularly fond of having all the textboxes stretch to the other side of the page, especially when some input don't need to be that long. Having said that, after visiting [http://webpublishing.anu.edu.au/web-style-guide/forms.php](url) it seems like ANU likes it so I'm ok with this.

cassdegrees/templates/managecourses.html
spencerj411: will do

cassdegrees/ui/views.py
spencerj411: will do

cassdegrees/templates/managecourses.html
JackSekoranja: So it looks like the ANU `tfull` CSS just makes the width of text boxes 70%. Maybe a solution is to create a custom CSS field doing the same thing for select boxes

cassdegrees/templates/managecourses.html
JackSekoranja: Alternatively maybe you could reposition year and units to be at the bottom and leave those smaller, while making the other text fields/select fields larger. Might need further discussion with team

cassdegrees/templates/managecourses.html
JackSekoranja: I still don't think we should restrict the software like that though. Maybe we could ask the user for confirmation if the entered year is too distant from the current year (Implementation may be too complicated for the resulting gain though)

spencerj411:
UI Screenshot after Jack's suggestions
image

spencerj411:
UI after implementing Jack's suggestions
image

JackSekoranja (approved):
Changes look good. There is an issue where the CSS for `<select>` objects look misaligned but not sure if we can do much about it.

cassdegrees/static/css/style.css
```diff
@@ -27,3 +27,11 @@
     justify-content: center;
     margin: 12px;
 }
+
+select.tfull {
+    width: 70.25%;
+}
+
+select.text{
+    border-radius: 3px;
```
j-selby, line 36: Not specific to this PR, but:

http://webpublishing.anu.edu.au/web-style-guide/style-sampler.php
`Please avoid using inline styles or creating custom style sheets`

We should be justifying each use of rules in this file.

cassdegrees/ui/views.py
```diff
@@ -95,3 +95,77 @@ def sampleform(request):
...
+    action = request.GET.get('action', 'Add')
+    if   action == 'Add'   : method = 'post'
+    elif action == 'Edit'  : method = 'patch'
+    elif action == 'Delete': method = 'delete'
+    else:
+         action =  'Add'   ; method = 'post'
+
+    action = 'Add'; method = 'post' # TODO Remove once other methods are implemented
+
+    courses = requests.get(request.build_absolute_uri('/api/model/course/?format=json')).json()
```
j-selby, line 108: Wait, is this request made even for POST requests, which don't appear to use 'courses' at all? This doesn't seem right.

j-selby (changes requested):

cassdegrees/ui/views.py
spencerj411: @JackSekoranja

cassdegrees/static/css/style.css
spencerj411: Not sure if it was mentioned above but it didn't seem like there was a `<select> `version of `tfull` in their css file hence the creation of new rules.

spencerj411:
also added a proper implementation of the error/success message when submitting a form (based off #59) using ANU's style guide.

image

cassdegrees/ui/views.py
```diff
@@ -95,3 +95,86 @@ def sampleform(request):
...
+
+            rest_api = requests.delete(model_api_url + id_to_delete + '/')
+
+            if rest_api.status_code == 204:
+                render_properties['msg'] = 'Course successfully deleted!'
+            else:
+                render_properties['is_error'] = True
+                render_properties['msg'] = \
+                    "Failed to delete course. An unknown error has occurred. Please try again."
+    return render(request, 'managecourses.html',
```
j-selby: Fix up spacing here?

cassdegrees/ui/views.py
```diff
@@ -95,3 +95,86 @@ def sampleform(request):
...
+    }
+    if request.method == 'POST':
+        # hard coded url; only temporary
+        model_api_url = request.build_absolute_uri('/api/model/course/')
+        post_data = request.POST
+        actual_request = post_data.get('_method')
+
+        # This method of transferring data to the API was inspired by:
+        # https://stackoverflow.com/questions/11663945/calling-a-rest-api-from-django-view
+        if actual_request == "post":
```
j-selby: We shouldn't be using this for these forms - we only need POST and GET. Get rid of this if statement entirely IMO.

cassdegrees/ui/views.py
j-selby: No longer relevant - main template is now always rendered.

cassdegrees/ui/views.py
Daniel-Jang: actual_request is now deleted.

Daniel-Jang (changes requested):
This form, as is, can't handle course prereqs properly as they are currently done (e.g. COMPXXXX and COMPXXXX or COMPXXXX).

Also, this needs to be tabbed with Edit, Add and Delete tabs with this spec.:

"When you click on manage courses, there are 3 tabs: Edit, Add and Delete. Screen defaults to edit courses tab, which would have the table of courses with a link that brings out a form that you can change and click submit.
Add tab would just be the Spencer's page in it's current form (look at most recent pic)
Delete would be like the edit page but instead the link submits a delete request to the API."

I have already had conversations with Jack and Spencer and they believe it is a good idea. Spencer has already started implementing this and I will be trying to help him so I can work on my user story: https://github.com/cass-degrees/CASS-Degrees-Code/issues/31

EDIT: Had a chat with Jack and we will have to have a discussion later with Aiden on how to store the requisites properly. Currently, there is no method of storing it properly. Hence, I will approve Spencer's ticket as there is nothing he can do in this moment.

Daniel-Jang (approved):

PR #55, opened 2019-04-11
Feature start creating subplan

joshcbee:
Now possible to navigate from home to the tool creator (entering identification details to be implemented separately).

PR #59, opened 2019-04-14
Implement basic degree creation workflow

### LINKED ISSUE #6: Start Creating Template (1)
As a CASS Staff member, 
I want to be able to start creating a new program template,
So that I can enter the related program details for Students to take advantage of. 
**Acceptance Criteria:**
- It is possible to navigate from the system’s homepage to the template creation tool

### LINKED ISSUE #17: Input Template Identification Details (3)
As a CASS Staff member, 
I want to insert the year, type (Masters, Bachelor, Diploma etc.), program code and title of the program template I am creating,
So that I can refer back to relevant data information for each program plan.

**Acceptance Criteria:**
- While editing a program template, the user can edit the year the program is for, the type of degree, the program title, and the program code
- If the year, and title or year and program code match a program already in the system, the user is notified that the current plan is potentially a duplicate

j-selby:
This PR adds a page for a CASS Staff member to be able to create degrees. Verification is included to ensure that pre-existing degrees are not overwritten, with an appropriate error message displayed if this is the case.

Resolves #6, resolves #17

EightXenon


JackSekoranja (changes requested):
I'm having a minor issue on the main menu where the 'Create Program Template' isn't centred, so it would be good if that could be fixed. Also, the first acceptance criteria for #6 was _"It is possible to navigate from the system’s homepage to the template creation tool"_. However, the current changes don't do this, as after adding a template it just gets put in to the database and the user is notified. To fix, just link to a blank page that will host the creation tool. I also checked creating a duplicate course and it worked perfectly. No other issues

j-selby:
`It is possible to navigate from the system’s homepage to the template creation tool`
I understood this to simply be navigating from the root page (/) to the creation page itself. Is this not the idea agreed on within e.g. https://github.com/cass-degrees/CASS-Degrees-Code/pull/55, or does this mean something different?

JackSekoranja:
I think your assumption is correct, although the 'template creation tool' was the page that allows you to add information about the degree requirements (Global restrictions, Majors, Courses, etc.). My issue was that after successfully adding a course the page does not change (i.e. you can still add new courses as opposed to being in the edit tool). What should happen is the ability to add courses is removed and you begin 'editing' the course you just added

j-selby:
@JackSekoranja I had in mind that it would be a singular page tool, rather then having different, inconsistent UIs for this stuff. Is this a bad idea?

JackSekoranja:
@j-selby Not a bad Idea, consistency is always something we should strive for, although it may make less sense depending on what you mean. If you are talking about a 'singular page tool' as something that is the same between Templates and Subplans, then the implementation may be a bit messy since there are slightly different rules for both. However, If you are talking about the 'singular page tool' as a shared page between the creation itself and the actual building of the rules then I think that's a good idea, but I think the creation tab should at least hide or become an edit tool for the story to be considered complete.

joshcbee (changes requested):
Can successfully add a new plan. Successfully rejected when code and year duplicated.

Currently fails criteria 2 for #17 - year and title
image

- Terminology should be consistent - e.g URL for user is currently 'create_degree' rather than 'create program'
- Suggest legend of form field changed from 'Add program template' to 'Program Details' to avoid duplication and build consistency across forms.
- Suggest changing error message to reflect nature of issue (e.g. program with the same code and year already exists.

j-selby:
`Currently fails criteria 2 for #17 - year and title`
Hmmm... Is this possible to enforce on the data model side of things?

`Terminology should be consistent - e.g URL for user is currently 'create_degree' rather than 'create program'`
OK, can do.

`However, If you are talking about the 'singular page tool' as a shared page between the creation itself and the actual building of the rules then I think that's a good idea, but I think the creation tab should at least hide or become an edit tool for the story to be considered complete.`
This was my understanding. I think making some fields in the creation tab read only (title + unit count can be changed freely I  guess) is much simpler then having different UIs for this.

j-selby:
Fixed homepage buttons for the changes in this PR, changed URL, rebased with `develop`.

JackSekoranja:
So the Year and Title issue can be fixed relatively easily. In my PR (#62) I added all the necessary imports that you'll need to check existence of the Year-Title relationship. To avoid merge issues I'd want it pushed first, but once it's done you could write something like `list(DegreeModel.objects.filter(name__iexact=new_name, year=int(year)).values())` and if that was empty then there wouldn't be any pre-existing relationship.

j-selby:
Year and name of program now verified for uniqueness. Error messages clarified (i.e to specify conflicts with name, code and year, if they occur). Rebased with `develop`.

cassdegrees/ui/views.py
```diff
@@ -158,6 +161,49 @@ def sampleform(request):
...
+        degree_dict = \
+            {
+                'code': post_data.get('code'),
+                'name': post_data.get('name'),
+                # Do some early validation of these fields
+                'year': int(post_data.get('year')),
+                'units': int(post_data.get('units')),
+                'degreeType': post_data.get('degreeType')
+            }
+
```
JackSekoranja, line 183:
```suggestion

        for k, v in degree_dict.items():
            render_properties[k] = v
            
```

Same changes as Josh, allows inputted values to appear in the form once it has been re-submitted

cassdegrees/templates/createprogram.html
```diff
@@ -0,0 +1,58 @@
...
+
+    <form class="anuform" action="" method="post">
+        {% csrf_token %}
+
+        <fieldset>
+            <legend>Configure Program Details</legend>
+
+            <p>
+                <label for="code">Program Code:</label>
+                <input class="text" id="code" type="text" name="code" aria-required="true" placeholder="e.g. BARTS" required>
```
JackSekoranja:
```suggestion
                <input class="text" id="code" type="text" name="code" aria-required="true" placeholder="e.g. BARTS" required value="{{ code|default:"" }}">
```

Makes the code value autofill once the page is sent

cassdegrees/templates/createprogram.html
```diff
@@ -0,0 +1,58 @@
...
+            </p>
+
+            <p>
+                <label for="name">Program Name:</label>
+                <input class="text" id="name" type="text" name="name" aria-required="true" placeholder="e.g. Bachelor of Arts" required>
```
JackSekoranja:
```suggestion
                <input class="text" id="name" type="text" name="name" aria-required="true" placeholder="e.g. Bachelor of Arts" required value="{{ name|default:"" }}">
```

Makes the name value autofill once the page is sent

cassdegrees/templates/createprogram.html
```diff
@@ -0,0 +1,58 @@
...
+            </p>
+
+            <p>
+                <label for="year">Program Year:</label>
+                <input class="text" id="year" type="number" min="2000" max="2100" value="2019" name="year" aria-required="true" required>
```
JackSekoranja:
```suggestion
                <input class="text" id="year" type="number" min="2000" max="2100" value="{{ year|default:"2019" }}" name="year" aria-required="true" required>
```

Makes the year value autofill once the page is sent

cassdegrees/templates/createprogram.html
```diff
@@ -0,0 +1,58 @@
...
+            </p>
+
+            <p>
+                <label for="units">Program Units:</label>
+                <input class="text" id="units" type="number" min="0" max="1000" value="144" name="units" aria-required="true" required>
```
JackSekoranja:
```suggestion
                <input class="text" id="units" type="number" min="0" max="1000" value="{{ units|default:"144" }}" name="units" aria-required="true" required>
```

Makes the units value autofill once the page is sent

cassdegrees/templates/createprogram.html
```diff
@@ -0,0 +1,58 @@
...
+            </p>
+
+            <p>
+                <label for="degreeType">Program Type:</label>
+                <select id="degreeType" name="degreeType" aria-required="true" required>
```
JackSekoranja:
```suggestion
                <select id="degreeType" name="degreeType" aria-required="true" required value="{{ degreeType|default:"" }}>
```

Makes the degree type autofill once the page is sent

JackSekoranja (changes requested):
Since you wanted to stay on the same page when adding the actual course components, I've suggested code changes that will re-fill the form correctly rather than deleting all the contents. Since the user story this completes has an acceptance criteria saying the user needs to be directed to the program creation page, we need to make sure the page reflects a proper creation tool after submitting, hence the need to keep the input values for the template. After implementing these changes I would be happy to approve the PR.

j-selby:
Actual editing is out of scope for this PR (kinda getting a bit big now), but I'll apply those anyway. Thanks!

JackSekoranja (approved):
Looks good, happy to merge.

joshcbee (approved):
Happy to merge.

PR #60, opened 2019-04-16
Feature input subplan identification details

joshcbee:
Allows user to add a subplan to the database by creating initial identification details. Does not redirect following creation at the moment as we need to determine whether further options to populate the subplan will appear below, to redirect to a new screen. Details persist on both success and failure to accommodate this future workflow.

**Acceptance Criteria:**
- While editing a subplan, the user can set the year the subplan is for, the title of the subplan, the code of the subplan, and whether it is a Major, Minor, or Specialisation
- If the year and title or year and code of the subplan match one already in the system, the user is notified that the current subplan is potentially a duplicate

image


cassdegrees/templates/createsubplan.html
```diff
@@ -6,6 +6,54 @@
...
+
+            <legend>Subplan details</legend>
+
+            <!-- As HTML forms only support get or post methods,
+            add this hidden input for consistency with the rest of the forms -->
+            <input type="hidden" name="_method" value="post">
+
+            <p>
+                <label for="code">Subplan Code:</label>
+                <input class="text" id="code" type="text" name="code"
```
JackSekoranja:
```suggestion
                <input class="text" id="code" type="text" name="code" placeholder="e.g. ANTH-MIN"
```

Recommend having placeholder text to be consistent with James

cassdegrees/templates/createsubplan.html
```diff
@@ -6,6 +6,54 @@
...
+                       aria-required="true" required value="{% if code != None %}{{code}}{% endif %}">
```
JackSekoranja:
```suggestion
                       aria-required="true" required value="{{ code|default:"" }}">
```

Simplified version based on  Django standard

cassdegrees/templates/createsubplan.html
```diff
@@ -6,6 +6,54 @@
...
+            </p>
+
+            <p>
+                <label for="year">Subplan Year:</label>
+                <input class="text" id="year" type="number" name="year"
+                       aria-required="true" required value="{% if year != None %}{{ year }}{% endif %}">
```
JackSekoranja:
```suggestion
                       aria-required="true" required min="2000" value="{{ year|default:"2019" }}">
```
Add lower bound on year. Set default year to 2019

cassdegrees/templates/createsubplan.html
```diff
@@ -6,6 +6,54 @@
...
+            </p>
+
+            <p>
+                <label for="name">Subplan Name:</label>
+                <input class="text" id="name" type="text" name="name"
+                       aria-required="true" required value="{% if name != None %}{{ name }}{% endif %}">
```
JackSekoranja:
```suggestion
                       aria-required="true" placeholder="e.g. Anthropology" required value="{{ name|default:"" }}">
```

Same as above

cassdegrees/templates/createsubplan.html
```diff
@@ -6,6 +6,54 @@
...
+            </p>
+
+            <p>
+                <label for="plantype">Subplan Type:</label>
+                <select class="text" id="planType" type="text" name="planType"
+                        aria-required="true" required value="{% if planType != None %}{{ planType }}{% endif %}">
```
JackSekoranja:
```suggestion
                        aria-required="true" required value="{{ planType|default:"" }}">
```
Same as above

cassdegrees/ui/views.py
```diff
@@ -95,8 +95,61 @@ def sampleform(request):
...
+        if rest_api.status_code == 201:
+            render_properties['msg'] = 'Subplan successfully added.'
+        else:
+            render_properties['is_error'] = True
+
+            rest_response = rest_api.json()
+            if "The fields code, year must make a unique set." in rest_response['non_field_errors']:
+                render_properties['msg'] = "A subplan already exists with this code and year."
+            elif "The fields year, name must make a unique set." in rest_response['non_field_errors']:
+                render_properties['msg'] = "A subplan already exists with this name and year."
```
JackSekoranja: I just realised that Majors and Minors can have the same name sometimes:
[ARAB-MAJ](https://programsandcourses.anu.edu.au/2019/major/ARAB-MAJ), [ARAB-MIN](https://programsandcourses.anu.edu.au/2019/minor/ARAB-MIN)

Might be a good idea to fix this. Either we should remove the `Name-Year` restriction or enforce a `Name-Year-Type` restriction

JackSekoranja (changes requested):
Recommending a few changes to ensure consistency between you and James. I did notice one problem with the user story, which is that different Subplans can have the same name and year. I've added details and suggestions below.

cassdegrees/ui/views.py
joshcbee: Have changed the restriction to Name-Year-Type for the moment to resolve this issue, probably preferable to removing the restriction altogether, but happy to revisit.

cassdegrees/templates/createsubplan.html
```diff
@@ -6,6 +6,54 @@
...
+            <p>
+                <label for="name">Subplan Name:</label>
+                <input class="text" id="name" type="text" name="name"
+                       aria-required="true" placeholder="e.g. Anthropology" required value="{{ name|default:"" }}">
+            </p>
+
+            <p>
+                <label for="plantype">Subplan Type:</label>
+                <select class="text" id="planType" type="text" name="planType"
+                        aria-required="true" required value="{{ planType|default:"" }}">
```
JackSekoranja:
```suggestion
                        aria-required="true" required">
```

Turns out adding `value=...` here doesn't work. See my suggested changes below

cassdegrees/templates/createsubplan.html
```diff
@@ -6,6 +6,54 @@
...
+                    <option value="MAJ">Major</option>
```
JackSekoranja:
```suggestion
                    <option value="MAJ" {% if planType == "MAJ" %}selected{% endif %}>Major</option>
```

Mark as selected if planType is MAJ

cassdegrees/templates/createsubplan.html
```diff
@@ -6,6 +6,54 @@
...
+                    <option value="MIN">Minor</option>
```
JackSekoranja:
```suggestion
                    <option value="MIN" {% if planType == "MIN" %}selected{% endif %}>Minor</option>
```

Mark as selected if planType is MIN

cassdegrees/templates/createsubplan.html
```diff
@@ -6,6 +6,54 @@
...
+                    <option value="SPEC">Specialisation</option>
```
JackSekoranja:
```suggestion
                    <option value="SPEC" {% if planType == "SPEC" %}selected{% endif %}>Specialisation</option>
```

Mark as selected if planType is SPEC

JackSekoranja (changes requested):
When testing I realised the subplan type doesn't keep its value after submission so I've suggested changes to fix it. @j-selby yours will probably suffer from the same issue :(

cassdegrees/ui/views.py
```diff
@@ -95,8 +95,64 @@ def sampleform(request):
...
+        if rest_api.status_code == 201:
+            render_properties['msg'] = 'Subplan successfully added.'
+        else:
+            render_properties['is_error'] = True
+
+            rest_response = rest_api.json()
+            if "The fields code, year must make a unique set." in rest_response['non_field_errors']:
+                render_properties['msg'] = "A subplan already exists with this code and year."
+
+                ""
```
j-selby, line 259: Not sure what this is doing here.

j-selby (approved):
Looks good!

JackSekoranja (approved):

Issue #29, opened 2019-04-03
Course Deletion (8)


JackSekoranja:
As a CASS Staff member, 
I want to be able to remove a course,
So that I can remove courses I have added by accident.

**Acceptance Criteria:**
- The user can select a course and is able to remove it from the system
- The course no longer appears in the system after deletion
- The user will be notified if the course is used in any subplans/templates, and will be asked to act on these before the deletion can occur

Issue #32, opened 2019-04-03
Template/Subplan Removal (8)


JackSekoranja:
As a CASS Staff member,
I want to be able to remove templates/subplans
So that I can ensure the templates/subplans on the system are accurate and up to date.

**Acceptance Criteria:**
- The user is able to select a template/subplan and remove it from the system
- The template/subplan no longer appears in the system after deletion
- In the case of a subplan deletion, the user will be notified if it is used in any templates, and will be asked to act on these before the deletion can occur

KavsHub:
I've been working on this story for the past few days and I've looked into the PR and code for #61 and #29.
It's pretty much a repeat (i.e copy and replace of course deletion to work also for programs and subplans).
I think we should generalise the manage_courses function to handle the requests for all the sections.
I'm going to have a chat with Spencer and Daniel about this as for now I've made changes which will conflict with their code later.

KavsHub:
Basic deletion has been completed in PR #70 - which has been merged.

To complete this story, we still need to complete the last acceptance criteria of:

- In the case of a subplan deletion, the user will be notified if it is used in any templates, and will be asked to act on these before the deletion can occur

KavsHub:
PR's #70 and #97 (which have been merged successfully)
Complete this story.

Closing this issue now.

PR #61, opened 2019-04-16
Base Code for Course Edit and Delete

spencerj411:
As #29 _Course Deletion (8)_ is currently blocked, I created a new branch that contains base code for both #31 _Course Editing (5)_ (@Daniel-Jang) and #29 instead.
The idea was to render just one instance of the course table and have an 'Options' column where the user could choose to either edit or delete a course. 

Notes:
1. code contains placeholders (sample links and images)
2. the icons used were from http://webpublishing.anu.edu.au/web-style-guide/icons.php. However, we are expected to contact ANU styling first if we want to use their icons for the MVP.
3. ANU styling has no icons that could be used for 'Delete' (hence the weird choice of thumbs down). However, we could also contact them about this if we want.

Capture

### DIFF
--- a/cassdegrees/templates/list.html
+++ b/cassdegrees/templates/list.html
@@ -35,7 +35,7 @@
             <form method="post">
                 {% csrf_token %}
                 {# Allow manage_courses() function to behave appropriately based on origin of request. The post request can either come from list.html or managecourses.html #}
-                <input type="hidden" id="request_origin" name="request_origin" value="list.html">
+                <input type="hidden" id="perform_function" name="perform_function" value="retrieve view from selected">
 
                 <table class="fullwidth tbl-cell-bdr">
                     {# Add title row based on what was in the dictionary #}
--- a/cassdegrees/templates/list.html
+++ b/cassdegrees/templates/list.html
@@ -32,35 +32,57 @@
     {# Generates table instances for each table in 'data', as well as the appropriate settings to keep it hidden #}
     {% for title, content in data.items %}
         <div id="{{ title }}Table" hidden>
-            <table class="fullwidth tbl-cell-bdr">
-                {# Add title row based on what was in the dictionary #}
-                <tr>
-                    {% for key in content.0 %}
-                        {% if key != "id" %}
-                            <th>{{ key }}</th>
-                        {% else %}
-                            <th>{{ title }}</th>
-                        {% endif %}
-                    {% endfor %}
-                </tr>
-                {# Add rows containing the contents of each json data instance #}
-                {% for row in content %}
+            <form method="post">
+                {% csrf_token %}
+                {# Allow manage_courses() function to behave appropriately based on origin of request. The post request can either come from list.html or managecourses.html #}
+                <input type="hidden" id="request_origin" name="request_origin" value="list.html">
+
+                <table class="fullwidth tbl-cell-bdr">
+                    {# Add title row based on what was in the dictionary #}
                     <tr>
-                        {% for key, item in row.items %}
-                            <td>
-                                {# If the item is in the ID column, generate a link to it instead #}
-                                {% if key != "id" %}
-                                    {{ item }}
-                                {% else %}
-                                    (<a href="/api/model/{{ title.lower }}/{{ item }}/">View</a>)
-                                {% endif %}
-                            </td>
+                        {% for key in content.0 %}
+                            {% if key != "id" %}
+                                <th>{{ key }}</th>
+                            {% else %}
+                                <th>Select</th>
+                                <th>{{ title }}</th>
+                            {% endif %}
                         {% endfor %}
                     </tr>
-                {% endfor %}
-            </table>
-            {# If there is no data available, notify the user #}
-            {% if not content %}<p>No {{ title }}s to display</p>{% endif %}
+                    {# Add rows containing the contents of each json data instance #}
+                    {% for row in content %}
+                        <tr>
+                            {% for key, item in row.items %}
+
+                                    {# If the item is in the ID column, generate a link to it instead #}
+                                    {% if key != "id" %}
+                                        <td>{{ item }}</td>
+                                    {% else %}
+                                        <td><input title="Select" type="radio" name="id" value="{{ item }}"  /></td>
+                                        <td>(<a href="/api/model/{{ title.lower }}/{{ item }}/">View</a>)</td>
+                                    {% endif %}
+
+                            {% endfor %}
+                        </tr>
+                    {% endfor %}
+                </table>
+                {# If there is no data available, notify the user #}
+                {% if not content %}
+                    <p>No {{ title }}s to display</p>
+                {% else %}
+                    <div class="right">
+                        <p class="left marginright">
+                           <input class="btn-uni-grad btn-large" type ="submit" formaction="/{% if title == 'Degree' %}{% elif title == 'Subplan' %}{% elif title == 'Course' %}manage_courses{% endif %}/?action=Add" value="Add a {{ title }}">
+                        </p>
+                        <p class="left marginright">
+                           <input class="btn-uni-grad btn-large" type ="submit" formaction="/{% if title == 'Degree' %}{% elif title == 'Subplan' %}{% elif title == 'Course' %}manage_courses{% endif %}/?action=Edit" value="Edit Selected">
+                        </p>
+                        <p class="right">
+                           <input class="btn-uni-grad btn-large" type ="submit" formaction="/{% if title == 'Degree' %}{% elif title == 'Subplan' %}{% elif title == 'Course' %}manage_courses{% endif %}/?action=Delete" value="Delete Selected">
+                        </p>
+                   </div>
+                </form>
+            {% endif %}
         </div>
     {% endfor %}
 
--- a/cassdegrees/templates/list.html
+++ b/cassdegrees/templates/list.html
@@ -81,8 +81,8 @@
                            <input class="btn-uni-grad btn-large" type ="submit" formaction="/{% if title == 'Degree' %}{% elif title == 'Subplan' %}{% elif title == 'Course' %}manage_courses{% endif %}/?action=Delete" value="Delete Selected">
                         </p>
                    </div>
-                </form>
-            {% endif %}
+                {% endif %}
+            </form>
         </div>
     {% endfor %}
 
--- a/cassdegrees/templates/managecourses.html
+++ b/cassdegrees/templates/managecourses.html
@@ -13,7 +13,7 @@
         <form class="anuform" action="/manage_courses/?action={{ action }}" method="post">
             {% csrf_token %}
             {# Allow manage_courses() function to behave appropriately based on origin of request. The post request can either come from list.html or managecourses.html #}
-            <input type="hidden" id="request_origin" name="request_origin" value="managecourses.html">
+            <input type="hidden" id="perform_function" name="perform_function" value="Add/Edit data to db">
             <input type="hidden" id="id" name="id" value="{{ render.edit_course_info.id }}">
             {% with required='<img src="//style.anu.edu.au/_anu/4/images/styles/asterisk.png" alt=" required">' %}
                 <fieldset>
--- a/cassdegrees/templates/managecourses.html
+++ b/cassdegrees/templates/managecourses.html
@@ -140,7 +140,6 @@
         }
 
         let tab = (new URLSearchParams(window.location.search)).get("action");
-        // when ?action=Modify, it gets sent as ?action=modify for some reason hence the fix below
         console.log(tab);
         let openElement = tab != null ? tab.charAt(0).toUpperCase() + tab.slice(1) : "";
         setActive(openElement);
--- a/cassdegrees/templates/managecourses.html
+++ b/cassdegrees/templates/managecourses.html
@@ -80,15 +80,23 @@
                     </p>
                 </form>
             {% else %} {# action == modify #}
-                <table class="fullwidth tbl-cell-bdr">
+                <div class=" nopadbottom tools-college fullwidth">
+                    <ul class="left">
+                        <li><a title="Tick one or more checkboxes" href="#">Mass Edit</a></li>
+                        <li><a title="Tick one or more checkboxes" href="#">Mass Delete</a></li>
+                    </ul>
+                </div>
+                <table class="fullwidth tbl-row-bdr noborder anu-long-area">
                     {# Add title row based on what was in the dictionary #}
-                    <tr>
+                    <tr class="anu-sticky-header">
                         {% for key in courses.0 %}
                             {% if key != "id" %}
                                 <th>{{ key }}</th>
+                            {% else %}
+                                <th><input title="Select all" type="checkbox" name="" value="" /></th>
                             {% endif %}
                         {% endfor %}
-                        <th>Options</th>
+                        <th>Actions</th>
                     </tr>
                     {# Add rows containing the contens of each json data instance #}
                     {% for row in courses %}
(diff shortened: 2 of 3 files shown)


JackSekoranja (changes requested):
I understand the reasoning behind using a 'thumbs down' picture, I don't think it's a good idea to have this in the final version. The layout of the pictures is also slightly different for me (See attached picture). I also think the newly created the newly created search function on the `/list/` page would be useful in this instance, and since they perform a very similar function, it may be better just to adapt that page and use it to find courses to edit/delete. Plus it means that we could have a consistent page for editing/removing existing courses, degrees, or subplans.

Edit Format

Also, just a note. You could continue to work on the actual deletion of courses themselves, as that will take some time. The only area you are blocked on is searching for course dependencies in subplans

spencerj411:
I should’ve been clearer about a few things:
1. The thumbs down icon was chosen temporarily just to see how 2 icons would look like in the Options column. We could either contact ANU styling about creating a new delete icon or we could just have them as text. Thoughts?
2. I have been working on the non-blocked criteria of Course Deletion but my thought process was to pr the overlapping work between delete and edit so @Daniel-Jang could start working on edit concurrently.

Idk why the icons are positioned like that though, I used their css; I’ll have a look into it. I’ll also have a look at `list.html` and implement the search function once changes have been merged.

JackSekoranja:
1. I think the preferred option here would be to find different pictures (outside ANU ones) for the edit and delete functions, or if we couldn't do that then just use text.
2. Fair enough

I also wasn't suggesting you implement the search function yourself, since it has a fair amount of overhead. I thought it might be easier to replace the edit form with the list form and have a single consistent one (Which would benefit code re-use). I was thinking we could separate the main menu in to 3-4 basic functions: 
- **Add New Template/Subplan/Course**
- **View Existing Template/Subplan/Course**
- **Edit Existing Template/Subplan/Course**
- **Output/Print Template**

Then the View/Edit forms could go to the link page, and from there they can re-direct to your specific edit or delete page

spencerj411:
I like the sound of that.

Daniel-Jang:
I'd also be happy to follow Jack's suggestion. What I also thought of is that we can remove the 'remove' button from the table list view and only have the edit button. The button can then lead to an edit page, which then instead has a button to delete the course or to commit new changes. 

That way, you don't have to worry about finding the right delete button.

Also, I reckon you should add text next to the edit icon ("Edit"), just so it is clear to the user.

spencerj411:
That sounds good to me too. But that has brought up another point which is what about if the user wants to mass delete a set of courses/templates/subplans?

spencerj411:
What about something similar to this? This way, it will take less clicks for the user to edit/delete. Also changed actions back to text, safer to have for the mean time.
image

JackSekoranja:
Yeah I was thinking something similar. It may even be better to remove the `edit and delete` text from the side and instead just have two buttons at the bottom saying edit selected and delete selected. I think this is a better idea since course editing and deletion is not supposed to happen very often, so it may be in our interests to make it MORE difficult to do

KavsHub:
Hey so I was looking at this and comparing to my issue of #32. 
In the long run we're going to have heaps of courses, subplans and templates which means the table would get very long. So future thinking/early brainstorming (both you and I should look into maybe multi-pages - i.e next page etc, or a search/filter box to help the user find the relevant stuff)

cassdegrees/ui/views.py
```diff
@@ -273,15 +273,21 @@ def manage_courses(request):
...
+            for id_to_delete in ids_to_delete:
+                rest_api = requests.delete(model_api_url + id_to_delete + '/')
 
-            if rest_api.status_code == 204:
-                render_properties['msg'] = 'Course successfully deleted!'
-            else:
+            if rest_api is None:
                 render_properties['is_error'] = True
-                render_properties['msg'] = "Failed to delete course. An unknown error has occurred. Please try again."
+                render_properties['msg'] = 'Choose a course to delete.'
```
j-selby: This is an exception, so this could be rephrased perhaps, or at least an exclamation mark?

cassdegrees/ui/views.py
```diff
@@ -231,6 +230,7 @@ def manage_courses(request):
         # actual_request = post_data.get('_method')
 
         if action == 'Add':
+            # courses = [{'code': course} for course in set([x['code'] for x in courses])]
```
j-selby: Is this here for a reason?

cassdegrees/templates/managecourses.html
```diff
@@ -79,15 +79,52 @@
...
+                       </p>
+                   </div>
+               </form>
             {% endif %}
         </div>
     {% endfor %}
     {# Function to hide the previous tab while opening a new one #}
     <script>
+
+
```
j-selby, line 26: Spacing here?

cassdegrees/ui/views.py
```diff
@@ -273,15 +273,21 @@ def manage_courses(request):
 
         # to be implemented, currently has the sample model code
         elif action == 'Delete':
-            id_to_delete = post_data.get('id')
-
-            rest_api = requests.delete(model_api_url + id_to_delete + '/')
+            ids_to_delete= post_data.getlist('id')
```
j-selby: Maybe verify these are integers? Else buggy code could potentially do unwanted stuff here (they are passed directly into the url below)

j-selby (changes requested):
No major complaints.

Daniel-Jang:
Uhh, James, you reviewed at an unfortunate time... I have completely rewritten everything...
Can everyone please check the most recent commit? It's hard to explain everything in words so yeah just run the program and see for yourselves :) (note: add, edit and delete functionality only works for courses at this stage)

With the most recent commit, I have also completed this ticket: https://github.com/cass-degrees/CASS-Degrees-Code/issues/31

cassdegrees/templates/managecourses.html
```diff
@@ -1,148 +1,76 @@
...
-                                    {% endfor %}
-                                </select>
-                            </p>
-                        </fieldset>
-                    {% endwith %}
-
-                    <p class="text-right">
-                        <input class="btn-uni-grad btn-large" type ="submit" value="{{ action }}">
+    <div id="{{ action }}Page" {% if action == 'Delete' or render.hide_form %}hidden{% endif %}>
+        <form class="anuform" action="/manage_courses/?action={{ action }}" method="post">
```
j-selby, line 13: We can't do this action stuff through POST?

cassdegrees/templates/managecourses.html
```diff
@@ -1,148 +1,76 @@
...
+            {% csrf_token %}
+            {# Allow manage_courses() function to behave appropriately based on origin of request. The post request can either come from list.html or managecourses.html #}
```
j-selby, line 15: IMO, this needs to not point to the origin url itself (there is a Referer header anyway, isn't there?), but as to the specific functionality that navigating from this page brings.

cassdegrees/ui/views.py
```diff
@@ -224,73 +223,100 @@ def manage_courses(request):
...
-                render_properties['msg'] = 'Course successfully added!'
-            else:
-                render_properties['is_error'] = True
-                # detects if the course already exists
-                if 'The fields code, year must make a unique set.' in rest_api.json()['non_field_errors']:
-                    render_properties['msg'] = "The course you are trying to create already exists!"
+        request_origin = post_data.get('request_origin')
+        print(request_origin)
+
+        if request_origin == 'managecourses.html':
```
j-selby: I think you need to add some documentation here about what the specific differences are for this. It wasn't immediately clear to me what this suggested.

cassdegrees/ui/views.py
```diff
@@ -224,73 +223,100 @@ def manage_courses(request):
...
-            rest_api = requests.post(model_api_url, data=course_instance)
-            if rest_api.status_code == 201:
-                render_properties['msg'] = 'Course successfully added!'
-            else:
-                render_properties['is_error'] = True
-                # detects if the course already exists
-                if 'The fields code, year must make a unique set.' in rest_api.json()['non_field_errors']:
-                    render_properties['msg'] = "The course you are trying to create already exists!"
+        request_origin = post_data.get('request_origin')
+        print(request_origin)
```
j-selby: Un-needed logging.

cassdegrees/ui/views.py
```diff
@@ -224,73 +223,100 @@ def manage_courses(request):
...
+                # Create a python dictionary with exactly the same fields as the model (in this case, CourseModel)
+                offered_sems = post_data.getlist('semesters[]')
+                course_instance = \
+                    {
+                        'code': post_data.get('code'),
+                        'year': post_data.get('year'),
+                        'name': post_data.get('name'),
+                        'units': post_data.get('units'),
+                        'offeredSem1': 'semester1' in offered_sems,
+                        'offeredSem2': 'semester2' in offered_sems
```
j-selby, line 299: Might be a good idea to note if there is something in offered_sems that *isn't* handled.

JackSekoranja (approved):
Looks good and works nicely. I would like to replace the radio buttons with check boxes so that the user can select and delete multiple courses at once, but I'm happy for us to create an issue to do that later so we can get this functionality out

cassdegrees/ui/views.py
```diff
@@ -224,73 +223,100 @@ def manage_courses(request):
...
-                    render_properties[
-                        'msg'] = "Failed to delete course. An unknown error has occurred. Please try again."
+                    render_properties['msg'] = 'Choose a course to delete.'
+                else:
+                    if rest_api.status_code == 204:
+                        render_properties['msg'] = 'Course successfully deleted!'
+                    else:
+                        render_properties['is_error'] = True
+                        render_properties[
+                            'msg'] = "Failed to delete course. An unknown error has occurred. Please try again."
```
j-selby: Not sure what happened here.

cassdegrees/ui/views.py
```diff
@@ -224,73 +223,100 @@ def manage_courses(request):
...
 
-            if rest_api.status_code == 200:
-                render_properties['msg'] = 'Course information successfully modified!'
-            else:
-                render_properties['is_error'] = True
-                render_properties['msg'] = "Failed to edit course information (unknown error). Please try again."
+                else:
+                    render_properties['is_error'] = True
+                    render_properties['hide_form'] = True
+                    render_properties['msg'] = "Choose a course to edit."
```
j-selby: Same feedback from before - rewording or at the very least an exclamation mark for an error.

cassdegrees/templates/list.html
```diff
@@ -32,35 +32,57 @@
...
+                        </p>
+                        <p class="left marginright">
+                           <input class="btn-uni-grad btn-large" type ="submit" formaction="/{% if title == 'Degree' %}{% elif title == 'Subplan' %}{% elif title == 'Course' %}manage_courses{% endif %}/?action=Edit" value="Edit Selected">
+                        </p>
+                        <p class="right">
+                           <input class="btn-uni-grad btn-large" type ="submit" formaction="/{% if title == 'Degree' %}{% elif title == 'Subplan' %}{% elif title == 'Course' %}manage_courses{% endif %}/?action=Delete" value="Delete Selected">
+                        </p>
+                   </div>
+                </form>
+            {% endif %}
```
j-selby: Wait, this endif ends *after* the form? That might break cases where there is no content.  (i.e `<form>` is unmatched).

j-selby (changes requested):

cassdegrees/templates/managecourses.html
Daniel-Jang: Which line are you talking about? I don't understand what you are suggesting.

cassdegrees/templates/managecourses.html
Daniel-Jang: What are your suggested changes? I'm not sure how to proceed from your comment.

cassdegrees/ui/views.py
Daniel-Jang: Added some comments explaining this.

cassdegrees/ui/views.py
Daniel-Jang: Forgot to remove them, now fixed.

cassdegrees/ui/views.py
Daniel-Jang: Not sure what you mean by something that isn't handled?

cassdegrees/ui/views.py
Daniel-Jang: Whoops, me neither, fixed now.

cassdegrees/ui/views.py
Daniel-Jang: Altered message to be "Please select a course to edit!"

cassdegrees/templates/list.html
Daniel-Jang: Fixed, didn't see that one.

cassdegrees/ui/views.py
Daniel-Jang: Changed message.

cassdegrees/ui/views.py
Daniel-Jang: No, don't think so, deleted.

cassdegrees/ui/views.py
Daniel-Jang: That shouldn't matter too much, the id's can't be altered by users (they are hidden input fields) and a post request with an unusual id made by something outside our Django app will be rejected since they don't have the required csrf token.

cassdegrees/templates/managecourses.html
j-selby: ` action="/manage_courses/?action={{ action }}"`
Why are we mixing GET and POST?

cassdegrees/templates/managecourses.html
j-selby: Instead of `managecourses.html`, name this option what it actually *changes*. If we are hardcoding the names of webpages here, this will get messy quick.

cassdegrees/ui/views.py
j-selby: offered_sems can potentially contain 'semester 1', 'semester 2', and (while we aren't explicitly handling it, as per our user stories), 'summer'/etc. Would just be nice to know if something unexpected is in here.

Daniel-Jang: Since 'semester 1' and 'semester 2' are defined by the form (and nothing else), there is no way we will get something unexpected. Users can't interact with this either. I think your suggested checks are overkill.

cassdegrees/templates/managecourses.html
Daniel-Jang: How are we mixing GET and POST? The form most certainly states that the method is post. Adding "?action=" is not an indicator that it is a GET request.

Also, while we can create a new hidden input with the action variable from managecourses.html, there is this one problem. From list.html, there is no way to determine which button the user will press on until they do and that 'submits' the form immediately. Therefore, there is no possible way to quickly change this hypothetical hidden input value to the one we want ("Add", "Edit" or "Delete") before the form is submitted. Hence, we will have to keep it this way.

cassdegrees/templates/managecourses.html
Daniel-Jang: Yeah OK that's a good reason. Will do.

cassdegrees/templates/managecourses.html
j-selby: Action in a URL is a [GET parameter](https://en.wikipedia.org/wiki/Query_string). Adding action in the URL there is by definition a query string, which is what is used for GET requests. This just seems a little weird to me.

Its not too big on an issue, but something to note.

j-selby (approved):

cassdegrees/templates/managecourses.html
Daniel-Jang: Actually, according to HTTP specifications, it is valid to have a query string for POST requests:
https://stackoverflow.com/questions/14710061/is-it-valid-to-combine-a-form-post-with-a-query-string/14710450
https://www.w3.org/Protocols/rfc2616/rfc2616.html

But I did just find out that it is generally not a good idea, so you were right to point that line out. Unfortunately, we still have to leave it this way due to the button problem :(

j-selby: Na, totally fair enough - just as long as this is noted, I personally have no issue (as you said, it isn't technically incorrect).

cassdegrees/ui/views.py
j-selby: Fair enough.

Issue #35, opened 2019-04-03
Template/Subplan Searching (13)


JackSekoranja:
As a CASS Staff member, 
I want to be able to search for an existing template/subplan,
So that I can easily view the plan when needed.

**Acceptance Criteria:**
- The user can search for plans using the template/subplan name, code, and year
- All matching results are returned
- No irrelevant results are returned
- The user can select the returned templates/subplans to start viewing/editing them

JackSekoranja:
Completed in PR #62

PR #62, opened 2019-04-16
Added search capabilities for degrees, subplans, and courses

JackSekoranja:
This PR is meant to close issue #35. I decided to also implement course searching (which was not in the original user story) so that others could take advantage of it if required. Note that you are only able to search using the year, code, and name for any given Degree/Subplan/Course. 
Search

### DIFF
--- a/cassdegrees/ui/views.py
+++ b/cassdegrees/ui/views.py
@@ -80,7 +80,7 @@ def planList(request):
 
         # If the degree, subplan, or course searches are non-empty, query the database
         data = {}
-        for target, model in [('Course', CourseModel), ('Subplan', SubplanModel), ('Degree', DegreeModel)]:
+        for target, model in [('Degree', DegreeModel), ('Subplan', SubplanModel), ('Course', CourseModel)]:
             # If the query is not blank, search for it in the database (Prevents unnecessary searches)
             if new_query[target]['AND'].children or new_query[target]['date'].children:
                 # SELECT from the the appropriate relation with the AND and OR queries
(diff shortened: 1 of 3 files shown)


cassdegrees/ui/views.py
```diff
@@ -31,11 +32,71 @@ def planList(request):
...
     """
-    degree = requests.get(request.build_absolute_uri('/api/model/degree/?format=json')).json()
-    subplan = requests.get(request.build_absolute_uri('/api/model/subplan/?format=json')).json()
-    course = requests.get(request.build_absolute_uri('/api/model/course/?format=json')).json()
+    query = request.GET.get('q', '')
 
-    return render(request, 'list.html', context={'data': {'Degree': degree, 'Subplan': subplan, 'Course': course}})
+    # No search, render default page
+    if not query:
+        degree = requests.get(request.build_absolute_uri('/api/model/degree/?format=json')).json()
```
j-selby, line 39: Is there a reason you have gone down the path of implementing search manually, rather than using a pre-built solution like https://fusejs.io/, which would likely do a few more things here (e.g. fuzzy search)?

Not that this is bad in itself though.

JackSekoranja: Admittedly I could have done a bit more research in to existing tools. Initially it was performing a much more basic search which wasn't too bad on the computation/technical front but the complexity grew once I decided to prioritise search results.

Daniel-Jang (approved):

cassdegrees/ui/views.py
j-selby: It doesn't particularly matter. if you do get time, fuzzy search would be nice here (i.e for typos), but other than that, LGTM!

j-selby (approved):

PR #63, opened 2019-04-16
Added Support for Degree Rules

aidengeorge97:
Degree rules have been mapped as a k-ary tree. See the [database details document](https://docs.google.com/document/d/13rDT9u6-vEgFtUdmtDeBNE_8HUAiiQryGiRG5CbEsxA/edit) or the class docstrings for information on how this is implemented. 

Integrating this pull will allow you to build expressions for DegreeRules and store them on the database. See issue #65 

Sample expression tree representation:

        "id": 3,
        "code": "BARTS",
        "year": 2019,
        "name": "Bachelor of Arts",
        "units": 144,
        "degreeType": "ugrad-sing",
        "rules": [
            "AND",
            [
                "OR",
                [
                    {
                        "code": "ANCH-MAJ",
                        "type": "subplan selection"
                    }
                ],
                [
                    {
                        "code": "HUEB-MAJ",
                        "type": "subplan selection"
                    }
                ],
                [
                    {
                        "code": "TNSO-MAJ",
                        "type": "subplan selection"
                    }
                ]
            ],
            [
                "OR",
                [
                    {
                        "code": "AUIS-MIN",
                        "type": "subplan selection"
                    }
                ],
                [
                    {
                        "code": "DEMO-MIN",
                        "type": "subplan selection"
                    }
                ],
                [
                    {
                        "code": "FORA-MIN",
                        "type": "subplan selection"
                    }
                ]
            ],
            [
                {
                    "type": "electives",
                    "units": 48
                }
            ]
        ]

### DIFF
--- a/cassdegrees/api/degreeRules.py
+++ b/cassdegrees/api/degreeRules.py
@@ -1,0 +1,52 @@
+class degreeRules:
+    JSONString = []
+
+    def getRule(self, ruleType, match):
+        rule = {"type": ruleType, "match": match}
+
+        return rule
+
+    """ 
+        ["(", A, "OR", "B", ")", "AND", "(", "C", "OR", "D", ")"]
+    """
+    def buildExpression(self, expressionList):
+        tree = {}
+        stack = []
+        node = tree
+        for token in expressionList:
+            if token == "(":
+                node['left'] = {}
+                stack.append(node)
+                node = node['left']
+            elif token == ")":
+                if len(stack) > 0:
+                    node = stack.pop()
+                else:
+                    print("h")
+                    parent = {}
+                    parent["left"] = node
+                    node = parent
+                    tree = parent
+            elif token in ["AND", "OR"]:
+                node['val'] = token
+                node['right'] = {}
+                stack.append(node)
+                node = node['right']
+            else:
+                node['rule'] = token
+                print(stack)
+                parent = stack.pop()
+                print(stack)
+                node = parent
+        return tree
+        #create rules, return rules. Then take a list of rules in a bool alg expression and build the json
+
+    # Add rule as another AND?
+
+
+
+degreeRule = degreeRules()
+A = degreeRule.getRule("single course", "COMP1***")
+test = ["(", A, "OR", "B", ")", "AND", "(", "C", "OR", "D", ")"]
+t = degreeRule.buildExpression(test)
+print(t)
--- a/cassdegrees/api/degreeRules.py
+++ b/cassdegrees/api/degreeRules.py
@@ -1,52 +1,161 @@
-class degreeRules:
-    JSONString = []
+import json
 
-    def getRule(self, ruleType, match):
-        rule = {"type": ruleType, "match": match}
-
-        return rule
-
-    """ 
-        ["(", A, "OR", "B", ")", "AND", "(", "C", "OR", "D", ")"]
+class DegreeRules:
     """
-    def buildExpression(self, expressionList):
-        tree = {}
+    A class which represents degree rules as a k-ary tree and stores them in a list.
+
+    Rules should be created using the rule_ functions, which return a dictionary
+    representing the rule. Rules can then be arranged into an expression and parsed
+    through the class to build their k-ary tree representation.
+
+    Calling get_rules_as_json will return the built expression as a JSON serialised string
+    that can be stored in the database under [degree].[rules].
+    """
+
+    def __init__(self, expression_list):
+        self.expressionTree = []
+        self.build_expression(expression_list)
+
+    def __init__(self):
+        self.expressionTree = []
+
+    def build_expression(self, expression_list):
+        """
+        Accepts a list of tokens that represent a boolean expression of degree rules
+        and converts them into a k-ary tree of lists and dictionaries that can be
+        stored in JSON format.
+
+        :param expression_list: (list) a list of tokens in the format of a boolean expression.
+                                A token must be one of:
+                                    "(", ")", "AND", "OR", rule
+                                where rule is some dictionary representing a rule, returned from a function below
+        :return: (list) a k-ary tree with rules as leaf nodes, and boolean operators as the other nodes
+        """
+
+        tree = [None]
         stack = []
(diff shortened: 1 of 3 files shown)


cassdegrees/api/degreeRules.py
```diff
@@ -1,52 +1,161 @@
...
+
+    def build_expression(self, expression_list):
+        """
+        Accepts a list of tokens that represent a boolean expression of degree rules
+        and converts them into a k-ary tree of lists and dictionaries that can be
+        stored in JSON format.
+
+        :param expression_list: (list) a list of tokens in the format of a boolean expression.
+                                A token must be one of:
+                                    "(", ")", "AND", "OR", rule
```
JackSekoranja, line 48: Just noting that it looks like this function requires the external tokens to be '(' and ')' as it will not work without them

cassdegrees/api/degreeRules.py
```diff
@@ -1,52 +1,161 @@
...
+
+        :param rule: (Dict) the rule to be appended
+        """
+
+        if self.expressionTree[0] == "AND":
+            self.expressionTree.append(rule)
+        else:
+            self.expressionTree = ["AND", self.expressionTree, rule]
+
+    def get_rules_as_json(self):
```
JackSekoranja, line 129: When generating rules for myself, the first rule was always `null`. Doesn't look intentional

cassdegrees/api/degreeRules.py
```diff
@@ -1,52 +1,161 @@
...
+    def rule_electives(units):
+        """
+        Creates a rule which describes a number of units worth
+        of elective courses which must be completed for the degree.
+
+        :param units: (Int) number of units of available electives
+        :return: (Dict) an elective rule
+        """
+
+        rule = {"type": "electives", "units": units}
```
JackSekoranja, line 177: This is fine for free electives, but there are no rules that allow you to generate list electives (24 units from courses A, B, C, D, E). We will need to have this in order to build more complex degrees

cassdegrees/api/degreeRules.py
```diff
@@ -1,52 +1,161 @@
...
+
+        :param subject: (String) a 4 letter course code i.e "ARTS", "COMP" etc.
+        :param level: (Int) a number in the range [1000, 2000, ... , 9000]
+        :param minimum: (Boolean) true = courses of the level or above count towards the rule,
+                                  false = only course equal to the level count towards the rule
+        :param units: (Int) number of units that must be completed from the subject area and level
+        :return: (Dict) a subject requirement rule
+        """
+
+        rule = {"type": "subject requirement", "subject": subject, "level": level, "minimum": minimum, "units": units}
```
JackSekoranja, line 195: This is useful for things like "No more than 24 units of 1000-level ARTS courses", but we may also want to represent restrictions/enforcements on the number of ARTS courses exclusively, as well as enforcements/restrictions on the number of year/level-specific courses exclusively. Function below has the same issue

JackSekoranja (changes requested):
Looks like we'll still need to make some changes to represent all rules. I'm not sure if the intention is for people to add rules to this list as they need them (which may be a good idea), but right now I think there is a lot missing

cassdegrees/api/degreeRules.py
aidengeorge97: Please elaborate? What were you doing to get a null?

cassdegrees/api/degreeRules.py
aidengeorge97: Valid. Do you want me to finalise all the degree rules before pulling this then? Because that's a full meeting with Deborah and extra requirement gathering. I want to pull this in now so the base functionality is there. Rules are easy to change after that.

cassdegrees/api/degreeRules.py
aidengeorge97: Yep, it's an incomplete set of rules. This is just the base functionality for storing rules and I fully expect many rules to be added as needed.

cassdegrees/api/degreeRules.py
aidengeorge97: Yes and no, but I found/fixed the problem you were pointing to anyway.

cassdegrees/api/degreeRules.py
JackSekoranja: I just wasn’t sure if this represented a finalised version. You’re probably right in thinking that we should update as we go so I’ll re-check my previous comments tomorrow (well today I guess) to see if they are fixed and will give you the go-ahead if they are

cassdegrees/api/degreeRules.py
JackSekoranja: @aidengeorge97 This is what I did to get `null`:
```
>>> from api.degreeRules import DegreeRules
>>> d_rules = DegreeRules()
>>> d_rules.build_expression([DegreeRules.rule_electives(24)])
>>> print(d_rules.get_rules_as_json())
[null, [{"type": "electives", "units": 24}]]
```

The same thing also happens with more complex rules such as `d_rules.build_expression(['(', DegreeRules.rule_electives(24), 'AND', DegreeRules.rule_electives(16), ')'])`

aidengeorge97: Ahhhhhh okay I didn't account for a scenario where a degree would only have one rule. Cause no degrees have one rule. But, valid, I'll just make it so one rule expressions default to an AND so it's a single term AND (i.e. just the one rule).

Not as certain about your second case though, I'll look into it.

aidengeorge97: Okay I solved the issue. Because you had brackets around those terms, it was expecting another set of rules, hence making a null parent node. I've added a function to flatten the tree, so those kinds of null nodes will be removed and their children get moved up. The logic was still right, but this will neaten the trees up and account for a wider range of poorly designed expressions. 

I also noticed an "issue" with append_rule while changing that, which is now fixed. Rather than just appending singular rules, you can now append a full expression list.

JackSekoranja (approved):
Okay, that looks better. Happy for you to merge now

Issue #28, opened 2019-04-03
Bulk Data Uploads (13)


JackSekoranja:
As a CASS Staff member, 
I want to be able to bulk add Majors, Minors, Specialisations, and Courses that have been given to me in document form (CSV, Spreadsheet, etc.),
So I don’t have to enter in each one manually when other data is available.

**Acceptance Criteria:**
- The user is able to upload a file from their computer and have the contents of the file be written to the database
- Adding the same file twice doesn’t create duplicate data
- Data that already exists in the system should be updated if it contains different information (eg. If a prerequisite changes for a course in a given year)
- The user can include semester information in course bulk upload.

PR #66, opened 2019-04-16
Bulk Data Upload

Daniel-Jang:
This feature allows courses and subplans to be added in bulk: https://github.com/cass-degrees/CASS-Degrees-Code/issues/28
image

Example Files here for reference:
[subplans.txt](https://github.com/cass-degrees/CASS-Degrees-Code/files/3085295/subplans.txt)
[courses.txt](https://github.com/cass-degrees/CASS-Degrees-Code/files/3085296/courses.txt)

### DIFF
--- a/cassdegrees/templates/bulkupload.html
+++ b/cassdegrees/templates/bulkupload.html
@@ -14,34 +14,34 @@
         </ul>
     </div>
 
+    {% if err_type != None %}
+        <p class="{% if err_type == 'error' %}msg-error{% elif err_type == 'warn' %}msg-warn{% elif err_type == 'success' %}msg-success{% endif %}">{{ user_msg }}</p>
+    {% endif %}
+
     {% for type in upload_type %}
         <div id="{{ type }}Page" hidden>
-            {% if type == 'Courses' %}
-                <form class="anuform" action="/bulk_upload/" method="post" enctype="multipart/form-data">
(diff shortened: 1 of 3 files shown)


j-selby:
I'm not sure we completely resolved this one - did we get info on what kinds of data the client might want to be uploading? e.g. format here is not csv but percent delimited.

JackSekoranja:
Yeah, I talked to Daniel about this and since the data we have seen didn’t have any structure it’s likely they will use it by generating the data specifically for the purpose of putting it in this system. That in addition to the fact that courses can have commas made us think that it was still reasonable to complete it as a TSV. Plus if this is available it opens the door to automation and other kinds of things - long story short as long as we can mass read in data in some form it’s still useful to the clients

Daniel-Jang:
Also James, this commit added explanations on why '%' was used instead of commas: https://github.com/cass-degrees/CASS-Degrees-Code/pull/66/commits/76e642527a05084ef9ffc48f6841e3c8ffc41a0c

Delimiter can be easily changed to something else if necessary
Jack did mention TSV since that's what I told him at the time and can be easily changed to that again.

JackSekoranja (approved):
Tested and looks good. I would have preferred you link it to the main menu somehow but I understand that may be outside of scope for the user story.

PR #67, opened 2019-04-18
Add support for adding/removing global requirements from a program

### LINKED ISSUE #15: Add Year-Specific Units Restriction (5)
As a CASS Staff member, 
I want to be able to create a ‘global requirement’ that restricts the number of year-specific (1000, 2000, 3000, etc.) units that are permitted by a program,
So students don’t get caught out by choosing more of these courses than their program allows.

**Acceptance Criteria:**
- While editing a program template, the user has the option of adding a restriction on the number of year-specific courses that can be used in a plan by inputting a number of units and then the target year
- Once this restriction has been added, the template has changed to indicate the presence of the restriction, clearly indicating the maximum units and the selected year

### LINKED ISSUE #16: Add Year-Specific Units Enforcement (5)
As a CASS Staff member, 
I want to be able to create a ‘global requirement’ that enforces the number of year-specific (1000, 2000, 3000, etc.) units that must be completed in a program,
So students don’t get caught out by choosing fewer of these courses than their program allows.

**Acceptance Criteria:**
- While editing a program template, the user has the option of adding a restriction on the minimum number of year-specific courses that can be used in a plan by inputting a number of units and then the target year
- Once this restriction has been added, the template has changed to indicate the presence of the restriction, clearly indicating the minimum units and the selected year

j-selby:
This is quite a big PR, so buckle in!

This PR adds support for specifying global requirements when creating a program. The intent is to re-use this entire page for editing a program.

Resolves #15. Resolves #16.

RobustWombat

FAQ:

- Why the custom serialization stuff?

While web browsers *technically* support lists in forms, they are not designed to group elements together (in, say, a global requirement or rule). Effectively, the work for this is now done on the client, rather than what would already have to be done on the server.

- Why a JSON field?

For the same reason as degree rules, this field allows for greater flexibility in storing the data, where a minor performance hit is a non-issue.

- Why the templating system?

Different kinds of global requirements/rules do need to be modeled throughout the life of the program, and this needed some degree of flexibility. We could have relied on the server constantly regenerating the server-side template for this, but that is poor UX for reasons of constantly having the page reload.

This solution, while requiring a bit more code, solves the UX dilemma.

- Why did I use Pristine?

Web browser's validation of forms was inconsistent for this, and didn't verify stuff like select boxes. Pristine was a 3 line addon to fix this and added proper messages. Alternatives required greater amounts of manipulation, or jQuery, which was way too overkill.

### DIFF
--- a/cassdegrees/templates/createprogram.html
+++ b/cassdegrees/templates/createprogram.html
@@ -39,7 +39,7 @@
 
             <p class="form-group">
                 <label for="degreeType">Program Type:</label>
-                <select id="degreeType" name="degreeType" aria-required="true" required value="{{ degreeType|default:"" }}>
+                <select id="degreeType" name="degreeType" aria-required="true" required value="{{ degreeType|default:"" }}">
                     <option value="ugrad-sing" selected>Undergraduate Single Pass Degree</option>
                     <option value="ugrad-doub">Undergraduate Flexible Double Degree</option>
                     <option value="hon">Honours Degree</option>
--- a/cassdegrees/templates/createprogram.html
+++ b/cassdegrees/templates/createprogram.html
@@ -1,4 +1,5 @@
 {% extends "base.html" %}
+{% load static %}
 
 {% block page-title %}Create Program Template{% endblock %}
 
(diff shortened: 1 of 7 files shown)


JackSekoranja (changes requested):
The only change I'd request you make is to hide the "Add Global Requirements" pane until the user actually creates a course. I'm open to discussion about this if you think it should be different, I just feel that it forces users to actually 'create' the plan first, which will remove the likelihood of people accidentally starting duplicate plans. 

I also think the remove requirement is a little awkward and it might be better to replace it with some kind of cross and/or reposition it, but for now I'm happy for you to leave it.

j-selby:
```
The only change I'd request you make is to hide the "Add Global Requirements" pane until the user actually creates a course. I'm open to discussion about this if you think it should be different, I just feel that it forces users to actually 'create' the plan first, which will remove the likelihood of people accidentally starting duplicate plans.
```
Yep! Can do. I'll leave this as-is until the edit page is linked in (so we don't accidentally break it in the meantime).

```
I also think the remove requirement is a little awkward and it might be better to replace it with some kind of cross and/or reposition it, but for now I'm happy for you to leave it.
```
Do agree. Not a bad idea - I'll have a brainstorm on it.

cassdegrees/templates/createprogram.html
```diff
@@ -1,44 +1,45 @@
...
+            <p class="form-group">
                 <label for="units">Program Units:</label>
                 <input class="text" id="units" type="number" min="0" max="1000" value="{{ units|default:"144" }}" name="units" aria-required="true" required>
             </p>
 
-            <p>
+            <p class="form-group">
                 <label for="degreeType">Program Type:</label>
-                <select id="degreeType" name="degreeType" aria-required="true" required value="{{ degreeType|default:"" }}>
+                <select id="degreeType" name="degreeType" aria-required="true" required value="{{ degreeType|default:"" }}">
```
JackSekoranja:
```suggestion
                <select id="degreeType" name="degreeType" aria-required="true" required>
```
Realised `value` here doesn't do anything

cassdegrees/templates/createprogram.html
```diff
@@ -1,44 +1,45 @@
...
                     <option value="ugrad-sing" selected>Undergraduate Single Pass Degree</option>
```
JackSekoranja:
```suggestion
                     <option value="ugrad-sing" {% if degreeType == "ugrad-sing" %}selected{% endif %}>Undergraduate Single Pass Degree</option>
```

Mark as selected if degreeType is ugrad-sing

cassdegrees/templates/createprogram.html
```diff
@@ -1,44 +1,45 @@
...
                     <option value="ugrad-doub">Undergraduate Flexible Double Degree</option>
```
JackSekoranja:
```suggestion
                     <option value="ugrad-doub" {% if degreeType == "ugrad-doub" %}selected{% endif %}>Undergraduate Flexible Double Degree</option>
```

Mark as selected if degreeType is ugrad-doub

cassdegrees/templates/createprogram.html
```diff
@@ -1,44 +1,45 @@
...
                     <option value="hon">Honours Degree</option>
```
JackSekoranja:
```suggestion
                     <option value="hon" {% if degreeType == "hon" %}selected{% endif %}>Honours Degree</option>
```

Mark as selected if degreeType is hon

JackSekoranja (changes requested):
I mentioned this before, but unless you make the following changes the degreeType field won't carry across

JackSekoranja (approved):
Looks good :)

j-selby:
Been a couple of days, so I will merge this one now (as previously agreed policy).

Issue #69, opened 2019-04-22
[Bug] add, edit and delete courses buttons do not appear when there are no courses.


KavsHub:
I did a fresh clone of the current merge and cleaned the database too.
What I noticed is that views does not have the add, edit and delete options if there are no courses in the database.

**No courses:**
<img width="1127" alt="no courses bug" src="https://user-images.githubusercontent.com/44992275/56496821-57631600-653e-11e9-9af5-dfaa5162274a.PNG">
**With 1 course (what we should see even with no courses)**
<img width="1128" alt="with courses" src="https://user-images.githubusercontent.com/44992275/56496822-57631600-653e-11e9-96d8-894dd30a2a90.PNG">

I have had a chat with Daniel and he said he would look into this when he is free from exams.

KavsHub:
I've fixed this issue in my branch.
Once its merged I'll close this issue.

KavsHub:
Fixed and merged

PR #70, opened 2019-04-22
Basic program and subplan removal + bug fixes

KavsHub:
So now in list view, you can add and delete programs and subplans using the relevant buttons.
It's pretty much an extension of the methods used by #61, except modified for programs and subplans
I've also fixed Bug #69 so once this PR merges I'll close the issue.

All acceptance criteria have been met except for:
- In the case of a subplan deletion, the user will be notified if it is used in any templates, and will be asked to act on these before the deletion can occur

@spencerj411 and I were going to collaborate on this since it is in both our deletion stories. 
Some of it is currently also being blocked as we don't have a way to add courses in subplans yet.

We also need to merge or discuss the program and subplan creation, as currently they're split from their relevant management sections and also editing needs to be completed.

### DIFF
--- a/cassdegrees/templates/list.html
+++ b/cassdegrees/templates/list.html
@@ -69,6 +69,11 @@
                 {# If there is no data available, notify the user #}
                 {% if not content %}
                     <p>No {{ title }}s to display</p>
+                    <div class="right">
+                        <p class="left marginright">
+                           <input class="btn-uni-grad btn-large" type ="submit" formaction="/{% if title == 'Degree' %}manage_programs{% elif title == 'Subplan' %}manage_subplans{% elif title == 'Course' %}manage_courses{% endif %}/?action=Add" value="Add a {{ title }}">
+                        </p>
+                    </div>
                 {% else %}
                     <div class="right">
                         <p class="left marginright">
--- a/cassdegrees/templates/list.html
+++ b/cassdegrees/templates/list.html
@@ -72,13 +72,13 @@
                 {% else %}
                     <div class="right">
                         <p class="left marginright">
-                           <input class="btn-uni-grad btn-large" type ="submit" formaction="/{% if title == 'Degree' %}{% elif title == 'Subplan' %}{% elif title == 'Course' %}manage_courses{% endif %}/?action=Add" value="Add a {{ title }}">
+                           <input class="btn-uni-grad btn-large" type ="submit" formaction="/{% if title == 'Degree' %}manage_programs{% elif title == 'Subplan' %}manage_subplans{% elif title == 'Course' %}manage_courses{% endif %}/?action=Add" value="Add a {{ title }}">
                         </p>
                         <p class="left marginright">
-                           <input class="btn-uni-grad btn-large" type ="submit" formaction="/{% if title == 'Degree' %}{% elif title == 'Subplan' %}{% elif title == 'Course' %}manage_courses{% endif %}/?action=Edit" value="Edit Selected">
+                           <input class="btn-uni-grad btn-large" type ="submit" formaction="/{% if title == 'Degree' %}manage_programs{% elif title == 'Subplan' %}manage_subplans{% elif title == 'Course' %}manage_courses{% endif %}/?action=Edit" value="Edit Selected">
                         </p>
                         <p class="right">
-                           <input class="btn-uni-grad btn-large" type ="submit" formaction="/{% if title == 'Degree' %}{% elif title == 'Subplan' %}{% elif title == 'Course' %}manage_courses{% endif %}/?action=Delete" value="Delete Selected">
+                           <input class="btn-uni-grad btn-large" type ="submit" formaction="/{% if title == 'Degree' %}manage_programs{% elif title == 'Subplan' %}manage_subplans{% elif title == 'Course' %}manage_courses{% endif %}/?action=Delete" value="Delete Selected">
                         </p>
                    </div>
                 {% endif %}
--- a/cassdegrees/templates/list.html
+++ b/cassdegrees/templates/list.html
@@ -60,10 +60,10 @@
                                     {% if key != "id" %}
                                         <td>{{ item }}</td>
                                     {% else %}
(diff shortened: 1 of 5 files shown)


KavsHub:
I also just realised that the selection doesn't work. That is, you can't select multiple programs/sub plans/courses using the radio buttons.

I'll see if I can fix it tomorrow. But this PR should be fine.
**EDIT** @spencerj411 told me that it's not meant to work since it was out of scope.
I'll still have a quick look at it and see if its doable.

KavsHub:
So I changed the button from radio buttons to checkboxes and since the code in views.py was made correctly we can now delete multiple if you tick the checkboxes. 

However, editing multiple only takes the last data point from the table. (If you select multiple)

cassdegrees/templates/list.html
```diff
@@ -69,16 +69,21 @@
                 {# If there is no data available, notify the user #}
                 {% if not content %}
                     <p>No {{ title }}s to display</p>
+                    <div class="right">
+                        <p class="left marginright">
+                           <input class="btn-uni-grad btn-large" type ="submit" formaction="/{% if title == 'Degree' %}manage_programs{% elif title == 'Subplan' %}manage_subplans{% elif title == 'Course' %}manage_courses{% endif %}/?action=Add" value="Add a {{ title }}">
```
spencerj411: I was thinking something along the lines of this.
1. user doesn't need to scroll down to add a course/sub-plan/program.
2. no duplicate code/ no need to worry about button alignment fo when there are/aren't any records to display.

I'm more than open to more suggestions.
image

KavsHub: This is pretty much what I did in my current uncommitted work. I'm happy to include this version since your reasoning is valid :)

spencerj411: we could also consider floating action buttons so no matter where you are on the page, you have convenient access to the buttons

KavsHub:
Okay, everything is working and fixed now. 

Happy to merge when everyone is happy.

cassdegrees/templates/list.html
```diff
@@ -58,29 +60,28 @@
                                     {% if key != "id" %}
                                         <td>{{ item }}</td>
                                     {% else %}
-                                        <td><input title="Select" type="radio" name="id" value="{{ item }}"  /></td>
-                                        <td>(<a href="/api/model/{{ title.lower }}/{{ item }}/">View</a>)</td>
+                                        <td><input title="Select" type="checkbox" name="id" value="{{ item }}"  /></td>
+                                        <td>
+                                            (<a href="/api/model/{{ title.lower }}/{{ item }}/">View</a>)
+                                            <input class="btn-uni-grad" name="id" type ="submit" formaction="/{% if title == 'Degree' %}manage_programs{% elif title == 'Subplan' %}manage_subplans{% elif title == 'Course' %}manage_courses{% endif %}/?action=Edit" value="Edit {{ title }} (ID: {{ item }})">
```
JackSekoranja:
```suggestion
                                            <input class="btn-uni-grad" name="id" type ="submit" formaction="/manage_{% if title == 'Degree' %}programs{% elif title == 'Subplan' %}subplans{% elif title == 'Course' %}courses{% endif %}/?action=Edit" value="Edit">
```

Code minimisation and styling changes

cassdegrees/templates/list.html
```diff
@@ -36,7 +36,9 @@
                 {% csrf_token %}
                 {# Allow manage_courses() function to behave appropriately based on origin of request. The post request can either come from list.html or managecourses.html #}
                 <input type="hidden" id="perform_function" name="perform_function" value="retrieve view from selected">
-
+                <p class="left">
+                    <input class="btn-uni-grad" type ="submit" formaction="/{% if title == 'Degree' %}manage_programs{% elif title == 'Subplan' %}manage_subplans{% elif title == 'Course' %}manage_courses{% endif %}/?action=Add" value="Add a {{ title }}">
```
JackSekoranja:
```suggestion
                    <input class="btn-uni-grad" type ="submit" formaction="/manage_{% if title == 'Degree' %}programs{% elif title == 'Subplan' %}subplans{% elif title == 'Course' %}courses{% endif %}/?action=Add" value="New {{ title }}">
```

Code minimisation and styling changes

cassdegrees/templates/list.html
```diff
@@ -58,29 +60,28 @@
                                     {% if key != "id" %}
                                         <td>{{ item }}</td>
                                     {% else %}
-                                        <td><input title="Select" type="radio" name="id" value="{{ item }}"  /></td>
-                                        <td>(<a href="/api/model/{{ title.lower }}/{{ item }}/">View</a>)</td>
+                                        <td><input title="Select" type="checkbox" name="id" value="{{ item }}"  /></td>
+                                        <td>
+                                            (<a href="/api/model/{{ title.lower }}/{{ item }}/">View</a>)
```
JackSekoranja, line 66: Would be good if this appeared as the button below, but I don't know how to do this while making it inline

JackSekoranja (approved):
I'd like it if the following styling changes were made, but it looks good to me

cassdegrees/templates/list.html
KavsHub: This is fine, but I kept ID there temporary so that the edit function worked. 
I need @Daniel-Jang to fix it since I couldn't find a way to parse ID over once the button was pressed. (ID is currently parsed by the checkbox being ticked - so if the checkbox isn't ticked the submit breaks because it gives the value of "edit" instead of "id"). Another problem is multi delete relies on the checkboxes id's (being ticked) so if I made another id function the delete function would break. 

I hope this explains what I'm trying to say, but I'll have a talk with Daniel about this.

cassdegrees/templates/list.html
KavsHub: This is fine I'll add it in

cassdegrees/templates/list.html
KavsHub: I'll have a look into it tomorrow.
But do you mean one column where they're on top of each other?
So:
| View |
| Edit  |

Issue #36, opened 2019-04-03
Template/Subplan Editing (8)


JackSekoranja:
As a CASS Staff member, 
I want to be able to edit an existing program template/subplan,
So that I can reflect any new changes in the actual program/subplan. 

**Acceptance Criteria:**
- It is possible to find and start editing all templates/subplans in the system
- Edited templates use the same interface as template/subplan creation

Issue #21, opened 2019-04-03
Template Notes Field (2)


JackSekoranja:
As a CASS Staff member
I can add a free text comment for students to a template
So I can convey additional information about the plan to students.

**Acceptance Criteria:**
- Users can enter comments in a template field.
- Text will be included as part of the printed template.

joshcbee:
Implemented by #100

Issue #20, opened 2019-04-03
Staff Notes Field (2)


JackSekoranja:
As a CASS Staff member
I can add a comment to a template in development that can be viewed by other staff
So that we discuss the plan’s development and reasoning

**Acceptance Criteria:**
- Users can enter comments in a template field.
- Users can view comments in a template field.

joshcbee:
Implemented by #100

PR #71, opened 2019-04-23
Feature template and subplan editing

joshcbee:
Implements #36, #21 and #20. Currently edit is accessed through the 'View' link in the Links/Manage courses pages, which now redirects to the editor rather than the api model.

**Acceptance Criteria**:

- It is possible to find and start editing all templates/subplans in the system
- Edited templates use the same interface as template/subplan creation

- Users can enter comments in a template field.
- Text will be included as part of the printed template.

- Users can enter comments in a template field.
- Users can view comments in a template field.

image
image

### DIFF
--- a/cassdegrees/templates/globalrequirementswidget.html
+++ b/cassdegrees/templates/globalrequirementswidget.html
@@ -1,59 +1,7 @@
-{% extends "base.html" %}
+<!DOCTYPE html>
+<html lang="en">
+
 {% load static %}
-
-{% block page-title %}Create Program Template{% endblock %}
-
-{% block subtitle %}Create Program Template{% endblock %}
-
-{% block content %}
-    {% if msg != None %}
-        <p class="{% if is_error %}msg-error{% else %}msg-success{% endif %}" id="serverError">{{ msg }}</p>
-    {% endif %}
-
-
-    <form class="anuform" id="mainForm" action="" novalidate method="post">
-        {% csrf_token %}
-
-        <fieldset>
-            <legend>Configure Program Details</legend>
-
-            <p class="form-group">
-                <label for="code">Program Code:</label>
-                <input class="text" id="code" type="text" name="code" aria-required="true" minlength="4" placeholder="e.g. BARTS" required value="{{ code|default:"" }}">
-            </p>
-
-            <p class="form-group">
-                <label for="name">Program Name:</label>
-                <input class="text" id="name" type="text" name="name" aria-required="true" minlength="5" placeholder="e.g. Bachelor of Arts" required value="{{ name|default:"" }}">
-            </p>
-
-            <p class="form-group">
-                <label for="year">Program Year:</label>
-                <input class="text" id="year" type="number" min="2000" max="2100" value="{{ year|default:"2019" }}" name="year" aria-required="true" required>
-            </p>
-
-            <p class="form-group">
-                <label for="units">Program Units:</label>
-                <input class="text" id="units" type="number" min="0" max="1000" value="{{ units|default:"144" }}" name="units" aria-required="true" required>
-            </p>
-
-            <p class="form-group">
-                <label for="degreeType">Program Type:</label>
-                <select id="degreeType" name="degreeType" aria-required="true" required>
-                    <option value="ugrad-sing" {% if degreeType == "ugrad-sing" or degreeType == None %}selected{% endif %}>Undergraduate Single Pass Degree</option>
-                    <option value="ugrad-doub" {% if degreeType == "ugrad-doub" %}selected{% endif %}>Undergraduate Flexible Double Degree</option>
-                    <option value="hon" {% if degreeType == "hon" %}selected{% endif %}>Honours Degree</option>
-                    <option value="mast-sing" {% if degreeType == "mast-sing" %}selected{% endif %}>Masters Single Degree</option>
-                    <option value="mast-adv" {% if degreeType == "mast-adv" %}selected{% endif %}>Masters (Advanced) Degree</option>
-                    <option value="mast-doub" {% if degreeType == "mast-doub" %}selected{% endif %}>Masters Flexible Double Degree</option>
-                    <option value="vert-doub" {% if degreeType == "vert-doub" %}selected{% endif %}>Vertical Flexible Double Degree</option>
-                </select>
-            </p>
-        </fieldset>
-
-        <!-- Everything after here is generated on the fly -->
-        <input class="text" hidden id="globalRequirements" name="globalRequirements">
-    </form>
 
     <!-- We still need a form for regular styling, though these will be serialized manually because of their list-like
          structure -->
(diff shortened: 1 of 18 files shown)


KavsHub:
Hey this looks good. It might conflict with my PR.
But now that editing is done I can add your sections to the todo edit sections I left empty. I'll see if I can integrate it in once either of our PR's merge.

cassdegrees/templates/list.html
```diff
@@ -59,7 +59,11 @@
                                         <td>{{ item }}</td>
                                     {% else %}
                                         <td><input title="Select" type="radio" name="id" value="{{ item }}"  /></td>
-                                        <td>(<a href="/api/model/{{ title.lower }}/{{ item }}/">View</a>)</td>
+                                        {% if title.lower == "degree" or title.lower == "subplan" %}
+                                            <td>(<a href="/edit/{{ title.lower }}/{{ item }}/">View</a>)</td>
```
j-selby: Not sure I particularly like this system of a single endpoint for degrees (should be programs) and subplans. They are not identical, and I feel that this will just lead to heaps of if statements in code.

cassdegrees/templates/globalrequirementswidget.html
```diff
@@ -0,0 +1,149 @@
...
+        </fieldset>
+    </div>
+
+    <script src="{% static 'js/pristine.js' %}" type="application/javascript"></script>
+    <script src="{% static 'js/programmanagement.js' %}" type="application/javascript"></script>
+    {% if globalRequirements != None %}
+        <script>
+            deserializeReqs({{ globalRequirements|safe }});
+        </script>
+    {% endif %}
```
j-selby: Newline.

cassdegrees/templates/globalrequirementswidget.html
```diff
@@ -0,0 +1,149 @@
+<!DOCTYPE html>
+<html lang="en">
```
j-selby: This is a widget - duplicate definition of HTML might break stuff

cassdegrees/ui/forms.py
```diff
@@ -0,0 +1,74 @@
...
+
+class EditProgramFormSnippet(ModelForm):
+    class Meta:
+        model = DegreeModel
+        fields = ('code', 'year', 'name', 'units', 'degreeType')
+        widgets = {
+            'code': forms.TextInput(attrs={'class': "text tfull"}),
+            'year': forms.TextInput(attrs={'class': "text tfull", 'type': "number"}),
+            'name': forms.TextInput(attrs={'class': "text tfull"}),
+            'units': forms.TextInput(attrs={'class': "text tfull", 'type': "number"}),
```
j-selby, line 14: This I *do* like, however. Can we adapt this to the entire project? Seems bad to have this duplicated through different means everywhere.

j-selby: Does this do validation of any kind/restrictions on editing code/etc?

cassdegrees/ui/views.py
```diff
@@ -207,6 +210,120 @@ def create_program(request):
...
+    if element == "degree":
+        elementDesc = "Program Template"
+        model_object = DegreeModel.objects.get(id=program_id)
+        object_dict = model_to_dict(model_object)
+        form = EditProgramFormSnippet(initial=object_dict)
+        comment_snippet = StaffNotesFormSnippet(initial=object_dict)
+        content_snippet = ProgramContentSnippet(initial=object_dict)
+
+        model_api_url = 'http://127.0.0.1:8000/api/model/degree/'
+        # print(model_to_dict(model_object))
```
j-selby: Leftover print statements shouldn't be here

j-selby (changes requested):
I really like what you have done with widgets here, though a few teething issues I suspect.

cassdegrees/templates/list.html
joshcbee: Happy to review this at a later date. It should be programs, but we need to go through and refactor the rest of our code to reflect this. At the moment, the List portion of develop generates a "degree" label, not a programs label.

In terms of endpoint, with good use of widgets, and single landing page for editing is viable, but it would be good to have a discussion tomorrow about how much difference there is actually going to be between programs and subplans (I have a feeling once they are fleshed out they will actually be rather similar).

cassdegrees/ui/forms.py
joshcbee: It can be adapted for anything with a model, including instantiating blank forms for creation. We can either add validation through django forms built in is_valid(), or overlaying javascript like you've done in other areas. Using this approach should hugely cut down duplication, and the associated html generation.

cassdegrees/ui/views.py
joshcbee: Will definitely remove prior to merge, just waiting to see if anyone ran into bugs first.

cassdegrees/templates/globalrequirementswidget.html
joshcbee: Thanks, wasn't sure about this one. Removed

cassdegrees/ui/forms.py
j-selby: Fair. This is a required feature of the underlying user stories for creation (and by extension, editing), so I would suggest trying Pristine or whatever just to clean this up - should be fairly trivial.

cassdegrees/templates/list.html
j-selby: We are going to need to have a cleanup of these endpoints at some point *anyway*, so we can have this discussion then, sure,

cassdegrees/ui/forms.py
joshcbee: At the moment it rejects all the required use cases with the exception of program(name, year) as this one doesn't have a uniqueness constraint in the DB. Again, something we should talk about tomorrow. Happy to implement further validation, but we should be taking a common approach across the board.

JackSekoranja (approved):
I just added a commit that will hide the global requirements, rules, and staff/student notes when viewing it in `/list`. Stylistically I would also prefer both notes sections to appear below the global degree rules, as we'd want to present the user with the most important information first. Looks good otherwise

joshcbee:
Agreed on the preferred styling. At the moment the limitation is that as global requirements is implemented as a separate form, it can't be integrated into another (and has to sit beneath).

KavsHub:
Hey since I merged my PR theres now going to be some conflicts to resolve.

I'll look into resolving this for you :)

cassdegrees/templates/list.html
KavsHub: This is going to cause a lot of issues with the current branch as we already have an edit button that gives an action of edit - which is better than going directly there.

In the merge, I've done for you I've left this out. Could you look into manage_subplans and manage_programs under the 'edit' sections and have your code there.

KavsHub (changes requested):
Please pull the merge I've done for you and look into integrating your edit sections into manage_programs and mange_subplans.

I'll discuss this with you during the meeting today.

j-selby:
Partially implemented now by #80 - comments/staff notes should be broken out.

KavsHub:
Can we reject and close this PR? It's very outdated now and we can create another branch and PR when this issue is resolved.

joshcbee:
Superceeded by forms refactoring and feature template notes field branches.

PR #72, opened 2019-04-23
Implement mechanism for defining rules within program templates

### LINKED ISSUE #7: Put Subplans In Program Templates (3)
As a CASS Staff member, 
I want to be able to insert an existing subplan to a program template,
So that It properly reflects to students the units required by the program.

**Acceptance Criteria:**
- When editing a program template, the user is able to put a subplan in to the template
- When adding a new subplan, the user is prompted with a interactive list of the subplans in the system, or is otherwise able to specify and choose any subplan
- Once a new subplan is added, the template has changed to indicate the presence of the added subplan
- User will be notified if attempting to add a subplan to the template more than once.

### LINKED ISSUE #9: Put Multiple M.E Subplans In Template (8)
As a CASS Staff member, 
I want to be able to insert a set of mutually exclusive subplans to a program template,
So that It properly reflects to students all the different options they have to graduate.

**Acceptance Criteria:**
- While editing a program template, the user is able to put multiple subplans in to the template at once, and have them all register as multiple program pathways
- When adding a new subplan, the user is prompted with a interactive list of the subplans in the system and can check multiple of these, or is otherwise able to specify and choose any set of subplans
- Once a new subplan is added, the template has changed to indicate the presence of the numerous subplans, and clearly conveys to the user that any one of the subplans should be chosen by the student
- Once a subplan has been added to the list, it cannot be added again

### LINKED ISSUE #12: Add Custom Text Units (2)
As a CASS Staff member, 
I want to be able to add units for a text-specified field (e.g. Exchange, internship),
So that I can represent coursework in the plan that isn’t in P&C.

**Acceptance Criteria:**
- While editing a program template, the user has the option of adding an unspecified course by inputting a number of units (that is divisible by 6) and some text to display in lieu of the course codes
- Once the new courses/electives are added, the template has changed to indicate the presence of the new information, and clearly displays the number of units and the entered text

j-selby:
This PR implements a mechanism for defining rules, as well as definitions for subplans and custom text units.

Resolves #7. Resolves #9. Resolves #12.

subplan-final

FAQ:

- Why vue.js?

Rules is a recursive data structure. Wrangling this in pure Javascript is going to lead to huge amounts of code written. With this solution, only a couple hundred lines are needed to completely implement this structure.

- Why vue.js in particular?

Other frameworks considered (e.g. Angular) require a significant component of tooling which is not productive to be introduced to this project. Further, vue can be used on a single element on a single page, rather than dictating the entire project's management of data.

- Why the framework as it is (rule_container/etc)?

This allows embedding of rules/groups of rules recursively, further down the track, as well as a degree of flexibility in implementing rules in the future without needing to reconfigure the entire system.

rule = a single rule
rule_container = a group of rules, as well as an add button

- Why the modal system?

For purposes of UX - custom styling was needed, as ANU doesn't provide anything like this. A modal allows us to explicitly ask the user a question, and for selecting the kind of rule that is desired, this works well here.

- Why both custom text units and subplans in a single PR?

Developing the rules framework required multiple kinds of rules to be implemented in order to test functionality/compatibility/serialization. To reduce complexity, this was done in parallel.

- How do I add my own rule?

Add a element to `COMPONENT_NAMES` at the top of rules.js, with the key being an internal variable name (i.e no spaces, eg. "course_list"), and the value being the text displayed to the reader for that component (e.g. "Course List").

In rules.js, implement a Vue component (examples can be taken from the pre-existing rules, fairly simple) with the name of "rule_" + your exact key from above.

Add a template in the main document (again, examples should be obvious here).

And you should be done - everything else will automatically work itself out (including serialization).


cassdegrees/static/js/rules.js
```diff
@@ -0,0 +1,180 @@
...
+            type: Object,
+
+            validator: function (value) {
+                // Ensure that the object has all the attributes we need
+                if (!value.hasOwnProperty("text")) {
+                    value.text = "";
+                }
+
+                if (!value.hasOwnProperty("units")) {
+                    value.units = 0;
```
JackSekoranja, line 97: Would be great if we could restrict the values to units of 6 here, as no other options are valid. Not sure if this is possible or how we would do it though

cassdegrees/static/js/rules.js
```diff
@@ -0,0 +1,180 @@
...
+    created: function() {
+        // Javascript has the best indirection...
+        var rule = this;
+
+        var request = new XMLHttpRequest();
+
+        request.addEventListener("load", function() {
+            rule.subplans = JSON.parse(request.response);
+        });
+        request.open("GET", "/api/model/subplan/?format=json");
```
JackSekoranja, line 42: Technically not every subplan should appear on the list since not all of them are directly interchangeable (i.e. major is 48 units while minors are 24 units). I might be able to modify my search query to be an API, and then you could just search for `-MAJ` or `-MIN` to get a list of majors and minors respectively. Let me know if this is something you'd like me to do (or you could do it yourself if desired)

JackSekoranja (approved):
I've added a few suggestions regarding the proposed implementation. Technically there are a few issues (most notably being able to select Majors and Minors as alternates to each other) but all of the acceptance tests for the user stories have been passed, so this can be something we can worry about later.

cassdegrees/static/js/rules.js
j-selby: This is already done - note the error in the screenshot. We don't actually force this (this would be difficult, but a warning appears when this isn't the case).

cassdegrees/static/js/rules.js
j-selby: Would just be as easy to do this clientside - some kind of warning perhaps when subplans of differing lengths are in the same selection?

Daniel-Jang (approved):

Issue #64, opened 2019-04-16
Link Subplans to Course Codes


aidengeorge97:
Currently in the model, Subplans are linked to unique courses, i.e. a course/year pair. We would like subplans to link to course codes that are irrespective of year. 

This should be changed in the model api.

aidengeorge97:
See #76 for details of how this was resolved.

PR #76, opened 2019-04-25
Link Subplans to Course Codes

aidengeorge97:
See issue #64. 

Unfortunately there is no way to directly link subplans to course codes as a foreign key, because course codes are neither unique nor are they a primary key. I see two options to navigating this, and the first of which I have implemented here.

1. Remove CoursesInSubplanModel and add an ArrayField for courses to the subplan model. Because there is no way to link Courses to Subplans through foreign keys, there is no point in having a ManyToMany through table, so CoursesInSubplanModel can be removed. The key impact of this change is that courses can be added to a subplan regardless of whether they exist or not. They're just represented by a string in a list. 
2. Change nothing in the database model, and put the onus on the front end to add all instances of a course to a subplan in CoursesInSubplanModel. I.e. when COMP2400 is added to the Databases Major, the front end should add (COMP2400, 2016), (COMP2400, 2017), (COMP2400, 2018), (COMP2400, 2019) to CoursesInSubplanModel. This is a messy option that bloats the table, and if (COMP2400, 2020) is added later, you would need to check every subplan that COMP2400 is already in, and add the 2020 version to that.

From here, courses can be added to subplans by appending to the relevant ArrayField in SubplanModel. I expect that users will be given a list of valid courses to choose from by pulling distinct values from the CourseModel.


JackSekoranja:
@aidengeorge97 I think a simple array field isn't enough to represent the rules for this since there are Subplans such as `24 units from X list and 12 units from Y list`. We may need to represent it as a list of rules similarly to how it was done for Programs/Templates, and then just parse that whenever we want to find out if courses are included in the subplan

aidengeorge97:
That's valid. I wasn't sure that we'd be including validation functionality for subplans beyond just checking if courses belong to their list. 

If that's what you want I'll just change the arrayField to a JSONField and it can be filled in as per program rules. This means the subplan add/edit functionality is gonna need a fair bit more expansion to accommodate the added complexity. Your call on whether I change it or leave it.

JackSekoranja:
Changing it sounds good to me, although it may be a good idea to hear other members' opinions first. I also don't think we are adding complexity here - we can't represent Subplan courses as a single list, and since Program creation will definitely require more complex rules to work IMO we'll actually be making it easier for ourselves due to code reuse.

aidengeorge97:
I believe this is ready to merge now. I made the suggested change to a JSON field. Also deleted the degreeRules class since that wasn't being used and it needed to go.

(Lowkey this PR is just deleting all the code I wrote that got overlapped)

j-selby (approved):
Looks good!

PR #79, opened 2019-04-28
Cleanup view management, use ModelForms for creating programs and subplans

j-selby:
This PR introduces a few changes:

- Views have been broken out into individual script files, improving code cleanliness + Git mergeability.
- Programs and subplan creation has been migrated to using Django's ModelForm system, solving a good majority of the validation + parsing problem. This also means that we no longer need to bounce around HTTP requests when managing these components of the application.
- Submission for programs has been streamlined, fixing issues with accessibility software where software would manually submit forms and Javascript would not activate.
- Labelling IDs for forms is correctly handled automatically


cassdegrees/ui/forms.py
```diff
@@ -0,0 +1,110 @@
...
+
+class EditProgramFormSnippet(ModelForm):
+    # Use custom handlers for JSON fields
+    globalRequirements = JSONField(field_id='globalRequirements', required=False)
+
+    class Meta:
+        model = DegreeModel
+        fields = ('code', 'year', 'name', 'units', 'degreeType', 'globalRequirements')
+        widgets = {
+            'code': forms.TextInput(attrs={'class': "text tfull"}),
```
JackSekoranja:
```suggestion
            'code': forms.TextInput(attrs={'class': "text tfull", 'placeholder': "e.g. BARTS"}),
```

Adds sample placeholder text to Program forms

cassdegrees/ui/forms.py
```diff
@@ -0,0 +1,110 @@
...
+            'year': forms.NumberInput(attrs={'class': "text tfull", 'type': "number"}),
```
JackSekoranja:
```suggestion
            'year': forms.NumberInput(attrs={'class': "text tfull", 'type': "number", 'max': 3000}),
```

Restrict max value in the input value itself (Note that 'min' doesn't work for some reason)

cassdegrees/ui/forms.py
```diff
@@ -0,0 +1,110 @@
...
+            'name': forms.TextInput(attrs={'class': "text tfull"}),
```
JackSekoranja:
```suggestion
            'name': forms.TextInput(attrs={'class': "text tfull", 'placeholder': "e.g. Bachelor of Arts"}),
```

Adds sample placeholder text to Program forms

cassdegrees/ui/forms.py
```diff
@@ -0,0 +1,110 @@
...
+            'units': forms.NumberInput(attrs={'class': "text tfull", 'type': "number"}),
```
JackSekoranja:
```suggestion
            'units': forms.NumberInput(attrs={'class': "text tfull", 'type': "number", 'step': 6, 'max': 512}),
```

Makes the form increase in steps of 6, and restricts the max value to 512

cassdegrees/ui/forms.py
```diff
@@ -0,0 +1,110 @@
...
+
+class EditSubplanFormSnippet(ModelForm):
+    # Automatically injected by default
+    units = forms.IntegerField(widget=forms.HiddenInput(), required=False)
+
+    class Meta:
+        model = SubplanModel
+        fields = ('code', 'year', 'name', 'units', 'planType')
+        widgets = {
+            'code': forms.TextInput(attrs={'class': "text tfull"}),
```
JackSekoranja:
```suggestion
            'code': forms.TextInput(attrs={'class': "text tfull", 'placeholder': "e.g. ANTH-MIN"}),
```

Adds sample placeholder text to Subplan forms

cassdegrees/ui/forms.py
```diff
@@ -0,0 +1,110 @@
...
+            'year': forms.NumberInput(attrs={'class': "text tfull", 'type': "number"}),
```
JackSekoranja:
```suggestion
            'year': forms.NumberInput(attrs={'class': "text tfull", 'type': "number", 'max': 3000}),
```

Restrict max value in the input value itself (Again, 'min' doesn't work)

cassdegrees/ui/forms.py
```diff
@@ -0,0 +1,110 @@
...
+            'name': forms.TextInput(attrs={'class': "text tfull"}),
```
JackSekoranja:
```suggestion
            'name': forms.TextInput(attrs={'class': "text tfull", 'placeholder': "e.g. Anthropology}),
```

Adds sample placeholder text to Subplan forms

JackSekoranja (changes requested):
Looks good, I've attached some suggestions related to how the form should appear to the users, including verification within the input itself. Also as an FYI, there are still forms for editing courses that do not use the same system (e.g. `http://127.0.0.1:8000/manage/subplans/?action=Edit&id=1`). Might be a good idea to change these now, or register that they need to be changed in the future.

Daniel-Jang (approved):
I've had a look and I do like what you have done with views.py (splitting it up). I have also had a look at forms.py, which I guess is the validation and I like the idea of a centralised validation section.

Overall, I have no major complaints and this pull request should benefit us greatly.

However, I did notice that I can create a program with only 5 units. I think there should be a check to make sure it's divisible by 6, like the one for course creation, but I still approved the changes since this feature isn't actually part of this pull request.

joshcbee (changes requested):
Looks good, only two minor changes:
- Program code validation in forms.py needs to be <3 rather than <4 (e.g. BIR)
- Can remove 'type': "number" where we have used. NumberInput

I'm happy to address @JackSekoranja's comment on editing when I add in the edit changes. My inclination would be to use the same templates as create, but with bound data, rather than use separate ones. Thoughts?

j-selby:
`Also as an FYI, there are still forms for editing courses that do not use the same system (e.g. http://127.0.0.1:8000/manage/subplans/?action=Edit&id=1).`
I've already done editing, I'll PR that one next (didn't want to have this one too big).

`My inclination would be to use the same templates as create, but with bound data, rather than use separate ones.`
Yes, this will be the case.

cassdegrees/ui/forms.py
j-selby: I'm not sure about double validation here - we already validate this serverside (see the clean_* methods). Thoughts?

cassdegrees/ui/forms.py
j-selby: I've leave min in there for clarify anyway (in case the web browsers get around to fixing it).

j-selby:
Comments addressed.

joshcbee (approved):
Looks good.

PR #80, opened 2019-04-29
Complete management tools for programs, subplans and courses

j-selby:
This moves programs, subplans and courses to use a complete creation,
edit and deletion solution based on ModelForms, based on work done
by Josh and others.

Endpoints are now configured as follows:
`/edit/<course/...>/`: Edit page for editing courses/.... id as query param.
`/create/<course/...>/`: Create page for creating new entries of courses/...
`/delete/<courses/...>/`: Create page for deleting a range of courses/etc. This takes an array of ids, and will display confirmation before deletion.

All of these endpoints will return to the listings page upon success, with a message displayed.

Create course page has been slightly edited:

create_course


JackSekoranja (approved):
Looks good, works perfectly. I like how you redirect to the list page after editing/adding things.

aidengeorge97 (approved):
This is good, and helps make the whole project more robust. I found no issues. It will be useful for me to have these changes finalised so I can take them into consideration as I implement views.

PR #82, opened 2019-04-29
Implement rule definitions w/ ModelForm system

### LINKED ISSUE #7: Put Subplans In Program Templates (3)

### LINKED ISSUE #9: Put Multiple M.E Subplans In Template (8)

### LINKED ISSUE #12: Add Custom Text Units (2)

j-selby:
This PR implements a mechanism for defining rules, as well as definitions for subplans and custom text units.

This commit also unified global requirements and rules, using the one Vue system to ease error checking + other factors.

Updated version of #72.

Resolves #7. Resolves #9. Resolves #12.

subplan-final

FAQ:

- Why vue.js?

Rules is a recursive data structure. Wrangling this in pure Javascript is going to lead to huge amounts of code written. With this solution, only a couple hundred lines are needed to completely implement this structure.

- Why vue.js in particular?

Other frameworks considered (e.g. Angular) require a significant component of tooling which is not productive to be introduced to this project. Further, vue can be used on a single element on a single page, rather than dictating the entire project's management of data.

- Why the framework as it is (rule_container/etc)?

This allows embedding of rules/groups of rules recursively, further down the track, as well as a degree of flexibility in implementing rules in the future without needing to reconfigure the entire system.

rule = a single rule
rule_container = a group of rules, as well as an add button

- Why the modal system?

For purposes of UX - custom styling was needed, as ANU doesn't provide anything like this. A modal allows us to explicitly ask the user a question, and for selecting the kind of rule that is desired, this works well here.

- Why both custom text units and subplans in a single PR?

Developing the rules framework required multiple kinds of rules to be implemented in order to test functionality/compatibility/serialization. To reduce complexity, this was done in parallel.

- How do I add my own rule?

Add a element to `COMPONENT_NAMES` at the top of rules.js, with the key being an internal variable name (i.e no spaces, eg. "course_list"), and the value being the text displayed to the reader for that component (e.g. "Course List").

In rules.js, implement a Vue component (examples can be taken from the pre-existing rules, fairly simple) with the name of "rule_" + your exact key from above.

Add a template in the main document (again, examples should be obvious here).

And you should be done - everything else will automatically work itself out (including serialization).


cassdegrees/static/js/rules.js
```diff
@@ -0,0 +1,226 @@
...
+        var rule = this;
+
+        var request = new XMLHttpRequest();
+
+        request.addEventListener("load", function() {
+            rule.subplans = JSON.parse(request.response);
+
+            rule.check_options();
+        });
+        request.open("GET", "/api/model/subplan/?format=json");
```
JackSekoranja, line 46: Might be a good idea to request from my API instead:
`/api/search/?select=id,code,name,year,units&from=subplan&year=subplan_year`

This would mean you could restrict by the year of the Subplan since staff won't be creating degrees for 2019 containing Subplans from other years (You'll obviously have to link `subplan_year` to the actual value). 

It might also be a good idea to add `&code=-MAJ`, `&code=-MIN`, or `&code=-SPEC` to search for only majors, minors, or specialisations, since they aren't supposed to be interchangeable. This would also require some changes to the UI to allow the user to select the plan type

JackSekoranja (approved):
I've added a suggestion below that would make it easy to restrict unnecessary results from the Subplan list. The `Add` and `Cancel` buttons are also slightly different from the ANU one, and it may be good to be consistent with those.

Since these are only relatively small changes, and many people are waiting on these changes to implement their stories, I'd be happy for this to be pushed now assuming we would be able to address these later.

cassdegrees/static/js/rules.js
j-selby:
```
Might be a good idea to request from my API instead:
/api/search/?select=id,code,name,year,units&from=subplan&year=subplan_year
```
Yeah, I'll take a look down the line, though it isn't critical for this PR.

```
It might also be a good idea to add &code=-MAJ, &code=-MIN, or &code=-SPEC to search for only majors, minors, or specialisations, since they aren't supposed to be interchangeable. This would also require some changes to the UI to allow the user to select the plan type
```
We already discussed this one. A warning does appear, and feel like this isn't all that required IMO. Proper search might be a better alternative.

j-selby:
```
The Add and Cancel buttons are also slightly different from the ANU one, and it may be good to be consistent with those.
```
ANU doesn't apply styling to `<button>`/`<a>` tags, even when they have the button CSS. I'll see how I can work around it.

Daniel-Jang (approved):
I have tested it out and it seems to behave well when creating programs and editing them (the pre-fill works well).

cassdegrees/static/js/rules.js
JackSekoranja: My issue wasn't that there is no warning, it's that we are letting them select mutliple mutually exclusive courses only to tell them they aren't allowed. So why let them select them in the first place? As you've said, the PR still works and passes all criteria, so I'm fine for you to push. Just thought it was something we could work on later

PR #86, opened 2019-04-29
Add a breadcrumb trail to pages

JackSekoranja:
These changes will add several clickable links to the top of the page that the user can click to navigate. This is specified easily within the any HTML file extending base.html by typing something similar to the following:
```
{% block breadcrumb-trail %}
    {% breadcrumb  '/page_link' 'Breadcrumb Title' %}
    {% endcrumb 'Non-Clickable Title' %}
{% endblock %}
```

I couldn't figure out how to easily link page urls without explicitly stating the endpoint. It is technically possible using `{% url list:index %}` within pure HTML, but not within Django template blocks (to my knowledge).

Home
Create Subplan


aidengeorge97 (approved):
Looks good, I tested it on each page and it works fine. Sensible structure.

Daniel-Jang:
I had a look and I noticed that whenever I click on either Create Program Template or Create Subplan from the home page, this appears:
image

That link leads to list.html, which wasn't used in this particular case to get to this page (from home -> Create Program Template).

Is this intentional? I wasn't sure if this was a design choice.

Apart from this, it looks good.

j-selby:
@Daniel-Jang I believe the angle here is that the home page is shortcuts, not direct components

cassdegrees/templates/createcourse.html
```diff
@@ -1,9 +1,20 @@
 {% extends "base.html" %}
+{% load breadcrumbs %}
 
 {% block page-title %}{% if edit %}Edit{% else %}Create{% endif %} Course{% endblock %}
 
 {% block subtitle %}{% if edit %}Edit{% else %}Create{% endif %} Course{% endblock %}
 
+{% block breadcrumb-trail %}
+    {% breadcrumb  '/list/?view=Course' 'Template/Subplan/Course' %}
```
j-selby: `Templates/...`
IMO this should be plural

j-selby (changes requested):
Looking good!

JackSekoranja:
@Daniel-Jang Yeah, it was kinda unavoidable since the breadcrumbs are generated on a per-page basis. I thought it was acceptable since there are `create new X` on the list page anyway. Plus if they want to get to the homepage they can just click the `home` icon

Daniel-Jang:
@j-selby @JackSekoranja Ah ok, that makes sense, thanks for clearing it up for me.

Daniel-Jang (approved):

cassdegrees/templates/createcourse.html
JackSekoranja: Actually this was done deliberately because I figured it made sense in terms to go `Program >> Add` instead of `Programs >> Add`. Happy to change it if you really wanted to but I wasn't sure which was better

JackSekoranja: I decided to add your changes after consultation with Daniel

Issue #18, opened 2019-04-03
Template Duplication (8)


JackSekoranja:
As a CASS Staff member,
I want to be able to make a new program template based on an existing program template,
So that I can modify an existing template to fit the new requirements.

**Acceptance Criteria:**
- The user is able to select a program template and has the option of duplicating it, which will send the user to the template creation tool and will automatically populate the new template with the fields in the original template

PR #87, opened 2019-04-29
Template Duplication

Daniel-Jang:
Original Issue: https://github.com/cass-degrees/CASS-Degrees-Code/issues/18

This allows the user to duplicate an existing degree/program that a user has selected from list.html for the user to alter it and create.


cassdegrees/ui/views/programs.py
```diff
@@ -14,7 +30,10 @@ def create_program(request):
             return redirect('/list/?view=Degree&msg=Successfully Added Program!')
 
     else:
-        form = EditProgramFormSnippet()
+        if duplicate:
```
j-selby, line 35: Wait, this will edit, not duplicate. You need to reset the ID, no?

j-selby (changes requested):

cassdegrees/templates/list.html
```diff
@@ -66,6 +66,10 @@
                                     {% else %}
                                         <td><p class="text-center"><input title="Select" type="checkbox" name="id" value="{{ item }}"  onchange="sessionStorage.setItem('selected', {{ item }})"/></p></td>
                                         <td>
+                                            {# Only Degree and Subplans will need this feature #}
+                                            {% if title == 'Degree' or title == 'Subplan'  %}
+                                                <a class="btn-uni-grad" href="/create/{% if title == 'Degree' %}program{% elif title == 'Subplan' %}subplan{% elif title == 'Course' %}course{% endif %}/?id={{ item }}&duplicate=True">Duplicate</a>
```
j-selby: Web syntax is generally lower-case false here.

cassdegrees/ui/views/programs.py
```diff
@@ -6,6 +6,22 @@
 
 
 def create_program(request):
+    duplicate = request.GET.get('duplicate', False)
```
j-selby: You are either returning a string or a boolean value here - little inconsistent.

cassdegrees/ui/views/programs.py
Daniel-Jang: No, I tried it out and it did not edit the existing one. That part of code is for GET requests, and that instance=instance was done just to display the contents of the program user selected to duplicate from.

I have left the POST part of create_program view function as it was before so adding is normal. Attempting to add a duplicate without changing anything will lead to this anyways:
image

cassdegrees/ui/views/programs.py
Daniel-Jang: Fixed it by setting it to 'false' string and then reassign duplicate variable as boolean True or False immediately after receiving it.

JackSekoranja (approved):
Looks good, works well. No complaints

KavsHub (approved):
This is pretty much the same as the subplan one. Happy for you to merge both of them in.

PR #91, opened 2019-04-30
Program reference consistency

JackSekoranja:
This PR removes the use of the words `Degree` and `Template`, and instead uses the word `Program`. 

I didn't go as far as renaming the project (which is called CassDegrees) as it doesn't really affect the rest of the project. Since the project was advertised to us as `cass degrees` I think this is fine.

Just a reminder that you will need to call the following commands to see changes:
`python manage.py makemigrations`
`python manage.py migrate`

NOTE: You will lose all of you Program data by implementing these changes


Daniel-Jang (approved):
Tested it out and there seems to be no problems.

Issue #34, opened 2019-04-03
View Existing Plans/Subplans (8)


JackSekoranja:
As a CASS Staff member, 
I want to be able to view the details of an existing plan/subplan
So that I can be aware of the details of that plan.

**Acceptance Criteria:**
- Created templates can be displayed in a human-readable format, and correctly represent all information that was inputted into the template

aidengeorge97:
Closed as of #94. View links now no longer direct to the api. Completion of this issue has spawned two maintenance issues #95 and #96, which focus on expanding view functionality as the project grows.

PR #94, opened 2019-05-01
Feature view plans and subplans

aidengeorge97:
See issue #34.

Rather than linking to the Django REST api, the view button from the list page now links to a pretty print display for showing programs, subplans, and courses. 

The printing currently supports all features available in the CASS Program Planner, but will need to be updated as new program and subplan rules are added, as well as when course requisites and incompatibilities are supported. 

image

image

image


KavsHub (approved):
Good stuff, I always didn't like how view just went to the rest API link.
Will be extra valuable for when we complete all the rules and electives for plans and sub plans

Daniel-Jang (approved):
Looking good from my end.

PR #97, opened 2019-05-01
Subplan deletion now checks if the subplan is used by a degree

KavsHub:
As the title says, subplan deletion now checks if it is used by any degrees before allowing the user to delete. This completes the final acceptance criteria of #32 
Works for if multiple are selected too. 
However, doesn't tell them which degree and subplan are affected. 
Wasn't sure how to convey this without it being a long error message. - Please tell me what you think and I'll update it before we merge

delete fail

### DIFF
--- a/cassdegrees/templates/list.html
+++ b/cassdegrees/templates/list.html
@@ -13,8 +13,8 @@
     {% if render.msg != None %}
         <p class="msg-success">{{ render.msg }}</p>
     {% endif %}
-    {% if render.errmsg != None %}
-        <p class="msg-error">{{ render.errmsg }}</p>
+    {% if render.error != None %}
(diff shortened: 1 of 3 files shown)


KavsHub:
UPDATE: 
Now notifies the user of which subplan code and program codes are affected. (Also works for multiple)
updated delete

spencerj411:
Looks good to me. As everyone during the team meeting yesterday agreed on blocking the delete, I will be reimplementing (again) my check for course deletion. In the error message, should we also include the year of which the subplans are from?

KavsHub:
I updated the formatting of the error message so it doesn't look cluttered.
image
I'm going to merge this now since it's done.

Issue #23, opened 2019-04-03
Put Compulsory/elective courses in subplan (8)


JackSekoranja:
As a CASS Staff member, 
I want to be able to insert a specified number of compulsory courses/List electives to a subplan,
So that It properly reflects to students the units required to complete that subplan.

**Acceptance Criteria:**
- While editing a subplan, the user has the option of adding a number of compulsory courses/List electives by inputting a number of units (that is divisible by 6) and then selecting a number of courses based on those stored in the system
- Users are able to search for specific courses using the course name or course code
- Once the new courses/electives are added, the template has changed to indicate the presence of these courses, clearly indicating either the included courses or the fact that they are free electives
- Once the user has entered a course, it cannot be entered again
- Users are notified if they attempt to add a course that doesn’t exist

Issue #11, opened 2019-04-03
Put Compulsory/elective courses in template (8)


JackSekoranja:
As a CASS Staff member, 
I want to be able to insert a specified number of compulsory/elective units to a degree template,
So that It properly reflects to students the units required by the program.

**Acceptance Criteria:**
- While editing a program template, the user has the option of adding a number of compulsory courses/list electives by inputting a number of units (that is divisible by 6) and then selecting a number of courses based on those stored in the system
- Users are able to search for specific courses using the course name or course code
- The user can also specify a number of free electives by inputting a number of units (that is divisible by 6)
- Once the new courses/electives are added, the template has changed to indicate the presence of these courses, clearly indicating either the included courses or the fact that they are free electives
- The user can’t submit a number of units that is less than the unit value of the list they are being added to, unless repeatable courses are present.
- A course can’t be added to the same list twice, but can be specified as repeatable.
- Users are notified if they attempt to add a course that doesn’t exist

PR #102, opened 2019-05-07
Feature subplan compulsory elective courses v2

Daniel-Jang:
Original Issue: https://github.com/cass-degrees/CASS-Degrees-Code/issues/23

Note that I list all available courses in the course selection drop-down menu, thus search related criteria is not applicable. I also had difficulty renaming everything from rules to 'requirements', to distinguish from rules in programs. Is this even desired? I am not sure.

image


Daniel-Jang:
Apologies in advance for inconsistent Javascript code :/

JackSekoranja (changes requested):
So I don't know why this wasn't changed sooner but the CASS staff actually need to be able to put in a number of courses who's units sum is less than the one entered, which goes directly against one of your acceptance criteria (Now fixed). This is because some courses are repeatable courses, and therefore if there is a repeatable course (Say, 6 units worth) and the specification says students have to complete 12 units, then the student needs to complete that course twice. Ironically, Techlauncher is one of these repeatable 6 unit courses.

Also, subplans should take course codes, not course IDs. This is because students can complete courses from any year, whereas IDs restrict to a single course-year instance

Daniel-Jang:
Edit: Ignore this comment, look at comment below for update. I am keeping this as evidence of discussion.

Actually, I do have to disagree on your suggestion of saving course codes instead of IDs. ANU Programs and Courses does not actually do this. For example, if you visit the 2016 version of Intelligent Systems Major (https://programsandcourses.anu.edu.au/2016/major/INTL-MAJ), you are brought to a 2016 version of the major. If you then go to "Study" then click on any courses (let's say COMP3620), you are brought specifically to the 2016 version of the course (https://programsandcourses.anu.edu.au/2016/course/COMP3620), not the current (2019) one.

This is a better way to handle this anyways, since courses can appear and disappear over time and not specifying year is going to cause problems on finding the right course page to bring up.

However, you are correct to point this out since my current implementation of being able to select courses from any year is incorrect. I can restrict the course selection to the specified subplan year only, like how the current ANU PnC does it.

Also, I will just remove the course unit count checking restrictions.

Daniel-Jang:
I had a chat with Jack an hour ago and I will proceed with Jack's suggestions. I have been convinced that we shouldn't be following PnC for the sake of following it. Therefore, disregard what I have said above.

Also, it was much easier to implement than expected. I expected 3-4 hours of work (a full re-implementation was thought to be needed), it actually took 15 minutes.

However, there is a bug where the course options in the drop-down selection is duplicated if the same course has multiple years. The duplicate checker checks for course codes, not id's, so selecting the same course codes from different years will still return the user with the duplication error.

JackSekoranja (approved):
Looks good, the aforementioned bug of showing duplicate courses has also now been fixed. Happy for you to merge

cassdegrees/static/js/requirements.js
```diff
@@ -0,0 +1,174 @@
+//! Vue.js based means of adding/removing requirements.
```
j-selby: This needs to be explicitly labeled as for subplans IMO. Is there a reason we are using a different set of scripts for this vs. programs? I don't see a reason why they can't mix and match (i.e have a Major depend on a Minor?)

j-selby (changes requested):

cassdegrees/static/js/requirements.js
Daniel-Jang: This file no longer exists since it is redundant. I have renamed requirements.html to subplanrules.html.

Daniel-Jang:
In my latest commit, I have managed to complete this user story too: https://github.com/cass-degrees/CASS-Degrees-Code/issues/11

Issue #85, opened 2019-04-29
Remake Subplan/Program Check in Delete Course


spencerj411:
Due to changes made in #76, the implementation for checking if a course is in any subplan/program before deleting a course must be changed. Completion of this issue will outdate #74 and will finally satisfy all criteria in #29.

PR #103, opened 2019-05-07
Bug Course Deletion Check v3000

spencerj411:
Aims to complete #29 and #85. Due to several changes in other files, I had to rewrite this criterion multiple times. Hopefully, it will be closed soon. I feel like this one is a tad tricky because there's quite a few combinations to take into account. But, essentially, the features of this pull request are:

- detects if one or more courses (CourseModel) are in one or more subplans

- detects if one or more programs contain one or more course (CourseModel), and/or one or more course codes (string),  and/or one or more subplans (SubplanModel)

- blocks the deletion completely if any of the above is true

- in the error message, the courses which are in the same subplan/program are grouped together onto one line

- in the error message, each subplan/program that is affected by the deletion are shown in a new line

- attempts to use @JackSekoranja's search api to its full potential

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,33 @@
+# CASS Degrees Code
+Primary Codebase for the TechLauncher CASS Degrees project
+
+For our landing page please go to: https://github.com/cass-degrees/CASS-Degrees
+
+## Our Git Branching Structure
+In this project we will have two main branches: **Master** and **Develop**
+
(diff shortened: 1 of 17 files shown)


JackSekoranja (approved):
Looks good, would be happy for you to merge

Daniel-Jang (approved):

PR #105, opened 2019-05-07
Fix Firefox being unable to submit programs

j-selby:
This fixes a bug where global requirements and rules were included in the main form, breaking form submission in specific circumstances for Firefox.

This does change the ordering in the form, but this shouldn't be an issue.


JackSekoranja (approved):
Really not a fan of having the student and staff notes above the rules field, but I'll assume this was done as part of the bug fix so I'll defer to your judgement there.

If possible it would be great to fix, but obviously functionality trumps styling

j-selby:
Yeah, we can't have it inline in the form there. Could do some Javascript to move the form around if needed.

PR #107, opened 2019-05-08
Make search API results distinct and other minor improvements

JackSekoranja:
It's also now possible to search for things exactly by searching
`?fieldname_exact=term`
e.g. (`id_exact=1` will only show 1, while `id=1` will also show results like 11)

### DIFF
--- a/cassdegrees/api/views.py
+++ b/cassdegrees/api/views.py
@@ -83,7 +83,9 @@ def search(request):
 
     # Generates a query mapping for "title":"containing text" relationships
     include = {x+"__icontains": request.GET.get(x, None) for x in columns if request.GET.get(x, None)}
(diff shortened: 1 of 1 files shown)


Daniel-Jang (approved):
Looks good to me.

cassdegrees/api/views.py
```diff
@@ -83,14 +83,15 @@ def search(request):
 
     # Generates a query mapping for "title":"containing text" relationships
     include = {x+"__icontains": request.GET.get(x, None) for x in columns if request.GET.get(x, None)}
+    include.update({x + "__iexact": request.GET.get(x+"_exact", None) for x in columns if request.GET.get(x+"_exact", None)})
```
JackSekoranja:
```suggestion
    include.update(
        {x + "__iexact": request.GET.get(x+"_exact", None) for x in columns if request.GET.get(x+"_exact", None)}
    )
```

Making suggested change to prevent pep-8 violation

PR #109, opened 2019-05-08
Suggested changes for PR #103.

JackSekoranja:
This will simplify the code and delete courses based on course codes rather than IDs in order to match the recent changes in PR #102

### DIFF
--- a/cassdegrees/ui/views/courses.py
+++ b/cassdegrees/ui/views/courses.py
@@ -37,92 +37,35 @@ def delete_course(request):
     gen_request.GET = {'select': 'id,code,year', 'from': 'course'}
     courses = json.loads(search(gen_request).content.decode())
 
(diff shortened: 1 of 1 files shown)


spencerj411:
Looks good to me. I've just added a few comments here and there. One thing that is probably worth considering is how to format the error message as it could potentially get too messy (eg. several courses which are in several subplans/programs more than once). But, it could be done later on. But yeah, it's much simpler than before when using id. It's unfortunate that it wasn't explicitly mentioned in specifications/we didn't decide earlier on how the course is supposed to be searched :(

spencerj411:
Another thing, does this mean #97 needs to be modified? If I can recall correctly, it also uses id for checking subplans. @KavsHub

spencerj411 (approved):
Review comments above ^

Issue #13, opened 2019-04-03
Put Year-Specific Units (5)


JackSekoranja:
As a CASS Staff member, 
I want to be able to insert a specified number of year-specific (1000, 2000, 3000, etc.) units to a degree template,
So that It properly reflects to students the units required by the program.

**Acceptance Criteria:**
- While editing a program template, the user has the option of adding a number of year-specific courses by inputting a number of units (that is divisible by 6) and then the target year for the courses (1000, 2000, 3000)
- Once this field has been added, the template has changed to indicate the presence of these courses, clearly indicating the number of units and the year of the courses

spencerj411:
closed - see #112

PR #112, opened 2019-05-09
Created Year-Specific Units Rule in Program Template Creation

spencerj411:
Aims to accomplish #13. The rule uses a textbox for the user to type in the course year level in (with a max string length constraint of 4). An error message will pop up if the user has entered a number that is not divisible by 1000 (similar to how the unit count is checked).
**whilst creating a template**
image
**after the template has been made**
image

### DIFF
--- a/cassdegrees/static/js/rules.js
+++ b/cassdegrees/static/js/rules.js
@@ -220,20 +220,6 @@ Vue.component('rule_year_level', {
             "redraw": false
         }
     },
-    created: function() {
-        // Javascript has the best indirection...
-        var rule = this;
-
-        var request = new XMLHttpRequest();
-
-        request.addEventListener("load", function() {
-            rule.courses = JSON.parse(request.response);
-
-            rule.check_options();
-        });
-        request.open("GET", "/api/search/?select=code,name&from=course");
-        request.send();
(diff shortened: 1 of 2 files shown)


spencerj411:
Update: changed the textbox to a dropdown menu (@Daniel-Jang's suggestion)
image

JackSekoranja (changes requested):
This doesn't work for me, as I only see a blank box when I choose to add new rules. However, it does work if I go back one commit, so the last commit (that changed to use a drop down menu) was the one that broke the code. I also think it would be better to call this menu `Level-specific Units`, but I'd be open to suggestions if anyone else had a better name for it

spencerj411:
Apologies, for some reason, not all modified files were pushed. It should work now (shouldn't see a blank widget).

JackSekoranja (changes requested):
It appears and I can enter in all information, but there is a bug when I save where the `year_level` field shows up as 'None', even when I have selected something. 

Also again, would be great if we could replace `Course Year-Level Specific Units` with ``Level-Specific Units

JackSekoranja (approved):
Looks good :). Happy for you to merge

Issue #114, opened 2019-05-09
Change "Edit" Button to "Save"


aidengeorge97:
When editing a course/subplan/program, the button to submit the changed form should say "Save" rather than "Edit".

spencerj411:
closed - see #133.

PR #133, opened 2019-05-09
Changed 'Edit' buttons to 'Save' buttons

spencerj411:
The title is self-explanatory. Acts upon feedback from UAT (see #114)
image
image


KavsHub:
This is good. Since this was requested by the Users and it makes sense to call it save too.
Happy for you to merge this feature update

KavsHub (approved):
Approved

Issue #129, opened 2019-05-09
Select All Option


aidengeorge97:
On the List page for viewing programs/courses/subplans, you can select items individually for deletion. It would be nice to be able to select all items that are currently shown on the table for a given search query, without having to individually select each one.

spencerj411:
closed - see #135.

PR #135, opened 2019-05-09
Implemented Select All Feature for Program/Subplan/Courses

spencerj411:
Aims to complete feature #129.
**Note:** Even with the use of anu styling, the alignment of the 'Select All' text and the checkbox within the same cell still looks a tad off. Either way, select all functionality works.
image

### DIFF
--- a/cassdegrees/templates/list.html
+++ b/cassdegrees/templates/list.html
@@ -59,10 +59,9 @@
                             {% if key != "id" %}
                                 <th>{{ key }}</th>
                             {% else %}
-                                <th>
-                                    <p class="text-center">
-                                        Select All<input title="Select All" type="checkbox" name="select_all" onclick="selectAll(this, '{{ title }}')">
-                                    </p>
+                                <th class="text-center">
+                                    Select All </br>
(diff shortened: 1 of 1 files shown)


cassdegrees/templates/list.html
```diff
@@ -59,7 +59,11 @@
                             {% if key != "id" %}
                                 <th>{{ key }}</th>
                             {% else %}
-                                <th>Select</th>
+                                <th>
+                                    <p class="text-center">
+                                        Select All<input title="Select All" type="checkbox" name="select_all" onclick="selectAll(this, '{{ title }}')">
```
KavsHub: I fixed your problem.
All you need to a break. 
You also don't even need the paragraph tag
```
<th class="text-center">
     Select All </br>
     <input title="Select All" type="checkbox" name="select_all" onclick="selectAll(this, '{{ title }}')">
</th>
```
image

KavsHub (changes requested):

KavsHub:
The select all feature is a weird request as you wouldn't really need to use it unless its for mass deleting which it isn't good for either haha. 
Regardless, good job just have a look at my change request.

KavsHub (approved):

Issue #136, opened 2019-05-10
Delete Initial commit file from the repo


KavsHub:
Can someone please delete the initial_commit.txt in their next Pull Request?
It shouldn't be here anymore.

KavsHub:
Completed by PR #137

PR #137, opened 2019-05-10
Deleted redundant initial commit file 

KavsHub:
This is for issue #136 since we keep forgetting to delete it we may as well delete the file formally.


Daniel-Jang (approved):
Approved

Issue #141, opened 2019-05-14
Bug - Deleting A Subplan


spencerj411:
When checking if a program uses a certain subplan before deleting it, it seems like the code assumes that every program has a subplan. I am able to reproduce the bug by selecting to delete a subplan that is not part of any program (throws a `KeyError` with exception value `ids`). When I select a subplan that is part of a program (when program rules has the `ids `field), it works.

KavsHub:
@spencerj411 are you sure? I can't seem to reproduce the bug. I made a subplan (fresh so not linked to anything) then I deleted it with no problems.

I'll double check the code and make it ignore checking if no programs use the subplan currently being deleted.

KavsHub:
Found the bug and fixed in PR #145

PR #145, opened 2019-05-14
Fixed bug in subplan deletion (Issue #141)

KavsHub:
A very easy fix to the issue where rules of a program didn't have any ids which would break the deletion check function. (Issue #141)
Simply checks if 'ids' key exists in rules for the program and if not then it won't check.


KavsHub:
Merging since its a really small fix.

PR #146, opened 2019-05-14
Fixed issue with number inputs taking non-numbers (Bug #132)

### LINKED ISSUE #132: Illegal Input "e" in Number Fields
Number fields currently allow users to input the letter "e". This is because they support scientific notation, but there is no reason for CASS Degrees to use this.

KavsHub:
This patch closes #132.

Number input boxes now won't allow e, + , - in them and only takes numbers.
You can, however, bypass it by copy and paste it in, but this is very unlikely to happen...
For now, this addresses the main issue of the user pressing e, -, + by accident rather purposely breaking it/ feeding it bad data.

### DIFF
--- a/cassdegrees/templates/widgets/rules.html
+++ b/cassdegrees/templates/widgets/rules.html
@@ -35,7 +35,7 @@
             <label>
                 Students must select...
             </label>
-            <input class="text" type="number" v-on:change="check_options" v-model="details.unit_count" min="0" step="6" max="1000" aria-required="true" required>
+            <input class="text" onkeydown="javascript: return event.keyCode === 8 || event.keyCode === 46 ? true : !isNaN(Number(event.key))" type="number" v-on:change="check_options" v-model="details.unit_count" min="0" step="6" max="1000" aria-required="true" required>
             units
         </p>
 
@@ -64,7 +64,7 @@
             <label>
                 Students must complete...
             </label>
-            <input class="text" type="number" v-on:change="check_options" v-model="details.unit_count" min="0" step="6" max="1000" aria-required="true" required>
+            <input class="text" onkeydown="javascript: return event.keyCode === 8 || event.keyCode === 46 ? true : !isNaN(Number(event.key))" type="number" v-on:change="check_options" v-model="details.unit_count" min="0" step="6" max="1000" aria-required="true" required>
             units
         </p>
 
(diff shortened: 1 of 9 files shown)


KavsHub:
Please let me know if I missed any number input boxes. I'm pretty sure I got them all.

j-selby:
There is alot of duplication here - can this be broken out into a separate Javascript function, or for this stuff to be checked at validation time maybe?

KavsHub:
So I tried to use a JavaScript file but for some reason I couldn't access the class or ID of the input tag as it's using a script template from vue. Not sure how the hierarchy is for it (if anyone knows would be great).
Which is why I just did the easier and faster option of adding a small snippet to each tag. 
The code is there so if anyone can get it working from an external JS file. That would be great as we can also work on adding the copy and pasting check too.(I have the code but it was too long to be viable for a quick tag insert)

cassdegrees/templates/widgets/rules.html
```diff
@@ -64,7 +64,7 @@
             <label>
                 Students must complete...
             </label>
-            <input class="text" type="number" v-on:change="check_options" v-model="details.unit_count" min="0" step="6" max="1000" aria-required="true" required>
+            <input class="text" onkeydown="javascript: return event.keyCode === 8 || event.keyCode === 46 ? true : !isNaN(Number(event.key))" type="number" v-on:change="check_options" v-model="details.unit_count" min="0" step="6" max="1000" aria-required="true" required>
```
j-selby: `javascript: return checkKeys(event)`

j-selby: and then somewhere else:
```javascript
function checkKeys(event) {
    return event.keyCode === 8 || event.keyCode === 46 ? true : !isNaN(Number(event.key));
}
```

j-selby:
see above.

KavsHub:
Moved the function to a javascript file (main.js) this file renames sample.js since we don't use it. feel free to put scripts that are called by a lot of pages in here. It's imported into base.html, alternatively if you feel like this is too much then I can just import it into the pages that require it but that ruins the idea of having a main.js file.

JackSekoranja (approved):
Looks good to me, happy for you to merge.

j-selby:
@KavsHub Think it should be handled on a per-page basis honestly - no need for this except for the create/edit pages themselves.

Issue #131, opened 2019-05-09
Notify User of Name Fields Containing Keywords like "Major" or "Minor" 


aidengeorge97:
There should be a somewhat consistent style guide for inputting plans. For example, if a user attempts to input a subplan with name "Archaeology Major", they should be alerted not to include "Major" in the name.

Issue #130, opened 2019-05-09
Unit Fields Should Increment and validate by 6


aidengeorge97:
All unit fields in the system should be restricted to increments of 6. This helps prevent users from mistaking number of courses with units.

KavsHub:
Pretty sure all the unit fields already increment by 6. An extension would be to check that the unit number entered is divisible by 6.

Issue #126, opened 2019-05-09
Course/Program Code Verification/Formatting


aidengeorge97:
Currently the Code field for courses/subplans/programs is largely unrestricted. It is possible for a user to input the same item twice if the codes are sufficiently different i.e. "COMP1100" and "comp1100". 

Some strict formatting needs to be enforced to prevent mass duplicates from being added to the system. 

Subplan codes should be the most challenging. Users have been shown to write "ANTH-MAJ" as "anth maj".

Issue #118, opened 2019-05-09
"Subplan Type" Label rather than "Plan Type" on the Subplan Form


aidengeorge97:
On the subplan edit and create forms, the label "Plan Type" should instead read "Subplan Type".

PR #148, opened 2019-05-14
Updated forms to have better input validation

KavsHub:
**EDIT** I've committed a lot of small UAT and issues into this PR (most were just 1 line changes) so this PR contains a lot of content changes.
This addresses the wishes of  #113, #118, #126, #130, #131 and a bug fix in subplan creation that I found along the way.  (#149 #101)
I've added a lot of strict validation that follows the standards from Programs and Courses.
Firstly all codes are now saved as uppercase, even if the user enters in lowercase or mixed.
All units must be divisible by 6.

**SUBPLANS**
Subplan now checks for the plan type in the course code
Capture
Subplan codes should have a '-' to split its plan type
Capture2
Subplans names shouldn't have the plan type as the code is made for that #131 
Capture3
I actually found this bug and fixed it, if nothing is selected we get a keyerror. I fixed this by catching the error and notifying the user to fill all fields
Capture4

**COURSES**
We now have very strict course codes which follow the current P&C syntax
If not 4 letters at the start
Capture5
if not 4 numbers after
Capture6
if code is too long or short
Capture7
I added this because codes are not all length 8, some codes have A, B C on them so I added an optional letter for it
Capture9

Please tell me what you think. I think this is good for ensuring accurate data and I've followed all the Programs and courses standards.

Also, let me know if theres any more strict checks you want me to do.


cassdegrees/ui/forms.py
```diff
@@ -151,9 +185,17 @@ class Meta:
 
     def clean_code(self):
         data = self.cleaned_data['code']
-        if len(data) < 4:
-            raise forms.ValidationError("This should be at least 4 characters!")
-        return data
+        print(len(data))
```
j-selby: Unneeded print statement

j-selby (changes requested):
Looks good.

cassdegrees/ui/forms.py
KavsHub: Thanks haha.

JackSekoranja (approved):
Looks good. I should mention that CASS currently offers a Major with the word 'Major' in the title ([Animation and Video Major](https://programsandcourses.anu.edu.au/2019/major/ANVI-MAJ)), so we wouldn't be able to add that to the system currently. However, I understand this was a specific UAT request, so it may be a good idea to bring this up next client meeting (and admittedly this seems to be the only Subplan with the Subplan type in the name).

KavsHub:
> Looks good. I should mention that CASS currently offers a Major with the word 'Major' in the title ([Animation and Video Major](https://programsandcourses.anu.edu.au/2019/major/ANVI-MAJ)), so we wouldn't be able to add that to the system currently. However, I understand this was a specific UAT request, so it may be a good idea to bring this up next client meeting (and admittedly this seems to be the only Subplan with the Subplan type in the name).

I feel like that was a CASS mistake haha

KavsHub:
This PR addresses 6 issues now since I've combined a lot of small UAT issues and other issues in the backlog which were mostly 1 line changes or small fixes. 
Going to merge this once I've tested it thoroughly

Issue #127, opened 2019-05-09
Sort List


aidengeorge97:
Users would like to be able to sort the main List of programs/courses/subplans by the fields. I.e. sort by Year to view latest courses, or sort by name to view courses in alphabetical/reverse alphabetical order.

JackSekoranja:
Closed with PR #152

PR #152, opened 2019-05-16
Add sort functionality to listings page

JackSekoranja:
Completes Issue #127 by making it possible to sort by field in the listings page by clicking the carets when hovering over a field.

These changes use list.js to implement the sorting, which can be found here: https://listjs.com/

List Sort

### DIFF
--- a/cassdegrees/templates/list.html
+++ b/cassdegrees/templates/list.html
@@ -1,4 +1,6 @@
 {% extends "base.html" %}
+{% load static %}
+
 {% load breadcrumbs %}
 
 {% block page-title %}Program/Subplan/Course List{% endblock %}
@@ -42,10 +44,11 @@
         </ul>
     </div>
 
+    <script src="{% static 'js/list.js' %}" type="application/javascript"></script>
     {# Generates table instances for each table in 'data', as well as the appropriate settings to keep it hidden #}
     {% for title, content in data.items %}
         <div id="{{ title }}Table" hidden>
-            <form method="post">
+            <form id="{{ title }}Sorter" method="post">
                 {% csrf_token %}
                 {# Allow manage_courses() function to behave appropriately based on origin of request. The post request can either come from list.html or managecourses.html #}
                 <input type="hidden" id="perform_function" name="perform_function" value="retrieve view from selected">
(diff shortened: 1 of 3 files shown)


Daniel-Jang (approved):
Sorting works well, no issues.

cassdegrees/templates/list.html
```diff
@@ -54,10 +57,11 @@
                 </p>
                 <table class="fullwidth tbl-cell-bdr">
                     {# Add title row based on what was in the dictionary #}
-                    <tr>
+                    <thead>
+                        <tr>
                         {% for key in content.0 %}
```
j-selby, line 62: Historically this would be indented.

cassdegrees/templates/list.html
```diff
@@ -101,6 +108,15 @@
                 {% endif %}
             </form>
         </div>
+
+        <script>
+            // Creates a list object to be used by list.js for sorting
+            var {{ title }}Options = {
+                valueNames: [{% for k,v in content.0.items %}'{{ title }}{{ k }}',{% endfor %}]
```
j-selby, line 116: Maybe commenting what this actually looks like after evaluation might be helpful?

cassdegrees/templates/list.html
```diff
@@ -101,6 +108,15 @@
...
+            };
+
+            var {{ title }}List = new List("{{ title }}Sorter", {{ title }}Options);
```
j-selby, line 119: This would break with spaces in titles, no?

j-selby (changes requested):
Looking good so far!

cassdegrees/templates/list.html
JackSekoranja: Not sure what you mean, it's already indented and I don't think a second indent after the comment is necessary

cassdegrees/templates/list.html
JackSekoranja: So `{{ title }}` is always a django relation name. To my knowledge, these can't have spaces, but even if they could there should only be "Program", "Subplan", "Course", so I don't think its such a huge issue. However, I can do it easily by going `{{ title|cut:" " }}`, but IMO this might make it less readable. Thoughts?

Issue #122, opened 2019-05-09
Add a Creation/Last Updated Date Field to Plans


aidengeorge97:
For each model, courses/subplans/programs, add a field (or two) that stores when the item was created and/or when it was last updated. This should be displayed in the table on the List page.

spencerj411:
done - see #153

PR #153, opened 2019-05-16
'Last Updated' Field to Course/Subplan/Program Models

spencerj411:
This pr aims to accomplish #122.
I've added a new field called `lastUpdated` per model which has a default value to the current date (using the datetime library). When an instance of a course/subplan/program is edited, it updates the value of `lastUpdated` to the current date. This can be seen on the listings page. 

**Note:** 
- I wasn't sure whether to add a second field for the purpose of remembering when the instance was created so I've left it for now.
- I set the default date to March 10 2019 during my first migration, hence some rows has it as its date (this has allowed me to check if the date gets updated when a successful edit occurs).
image

### DIFF
--- a/cassdegrees/api/models.py
+++ b/cassdegrees/api/models.py
@@ -16,7 +16,7 @@ class CourseModel(models.Model):
     units = models.PositiveIntegerField()
     offeredSem1 = models.BooleanField()
     offeredSem2 = models.BooleanField()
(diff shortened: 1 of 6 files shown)


Daniel-Jang:
I reckon it might be a good idea to use Django's default, as suggested here:
image

Apart from this, I'm happy with the functionality.

Daniel-Jang (approved):
Yeah just look at my suggestions and then merge.

JackSekoranja (approved):
Looks good, but I also agree with Daniel's suggestions

PR #154, opened 2019-05-17
Finalised program Either Or rules

Daniel-Jang:
Either Or clauses can now be added to programs:
image

### DIFF
--- a/cassdegrees/static/js/rules.js
+++ b/cassdegrees/static/js/rules.js
@@ -357,7 +357,7 @@ Vue.component('rule_custom_text', {
 
 Vue.component('rule_either_or', {
     props: {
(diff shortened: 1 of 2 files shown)


JackSekoranja (changes requested):
It works well, but there are some valid rules that we still can't represent using this logic.

For example, `Either X or Y or Z`:
https://programsandcourses.anu.edu.au/2019/program/BARTS

Daniel-Jang:
Wow really? Searched over 10 courses and not a single one I checked had more than one or statement... Bad luck I guess.

JackSekoranja (approved):
Looks good to me, happy for you to merge

PR #155, opened 2019-05-21
Feature authentication

JackSekoranja:
Added logins to the system and prevented access to all UI elements without an authorised account. I wasn't able to remove external get access to our database as some of our components call it externally, but external post requests have been disabled.

Login
AccountsPage
PasswordChange
PasswordChangeSuccess


cassdegrees/api/Validators.py
```diff
@@ -0,0 +1,30 @@
...
+        if count < 3:
+            raise ValidationError(   
+                _("The password must contain at least 3 of the following classes: lowercase characters, uppercase characters, digits, punctuation, unicode characters"),
+                code='password_anu_fail',
+            )
+
+    def get_help_text(self):
+        return _(
+            "The password must contain at least 3 of the following classes: lowercase characters, uppercase characters, digits, punctuation, unicode characters"
+        )
```
j-selby: Newline

cassdegrees/cassdegrees/settings.py
```diff
@@ -102,12 +107,13 @@
     },
     {
         'NAME': 'django.contrib.auth.password_validation.MinimumLengthValidator',
+        'OPTIONS': {'min_length': 10}
     },
     {
         'NAME': 'django.contrib.auth.password_validation.CommonPasswordValidator',
```
j-selby: Are these other validators needed?

cassdegrees/templates/base.html
```diff
@@ -81,4 +95,4 @@ <h1 class="title" id="page-title">{% block subtitle %}No title defined{% endbloc
         </div>
     </div>
 </body>
-</html>
+</html>
```
j-selby: Removed newline

cassdegrees/templates/registration/login.html
```diff
@@ -0,0 +1,33 @@
...
+                </div>
+            {% endif %}
+            {# Render form fields #}
+            {% for field in form.visible_fields %}
+                <p>{{ field.label_tag }} {{ field }}</p>
+            {% endfor %}
+        </fieldset>
+
+        <p class="text-right">
+            <a class="left" onclick="alert('Please contact CASS administrators to change your password')">Forgot Password?</a>
```
j-selby: Onclick on a isn't a good idea. Either use href or a `<button>`

cassdegrees/templates/registration/password_change_done.html
```diff
@@ -0,0 +1,16 @@
...
+{% block subtitle %}CASS Staff Password Change{% endblock %}
+
+{% block breadcrumb-trail %}
+    {% finalcrumb 'Change Password' %}
+{% endblock %}
+
+{% block content %}
+    <div class="msg-success">Your password has been successfully changed</div>
+    <br />
+    <a href="/">Back to Hompage</a>
```
j-selby, line 15: Homepage

j-selby (changes requested):
Looking good. Few minor comments.

`external post requests have been disabled`
Nice

cassdegrees/cassdegrees/settings.py
JackSekoranja: Technically no, but some of them ensure the user doesn't enter in a password that has their username in it/other things like that. It's basically to make an attempt at following the privacy standards found here: https://policies.anu.edu.au/ppl/document/ANUP_013008

Specifically, it mentions the following:
- the account name in any form (as-is, reversed, capitalised, doubled, etc.);
- the user’s first or last name in any form;
- simple patterns of letters on keyboards; or
- any well-known or publicly posted identifiable information.

Technically all of these aren't checked for, but I think this is partially a protocol for users to follow so it might not be a huge issue. Having said that, I'll try and remove some of the ones that aren't strictly necessary

cassdegrees/templates/registration/password_change_done.html
JackSekoranja: Not sure what you are requesting here.

cassdegrees/templates/registration/login.html
JackSekoranja: I'll look in to it, but this seemed like a nice way of quickly notifying the user

cassdegrees/cassdegrees/settings.py
j-selby: Does this also apply to authentication then? If a weak password is setup for a user for whatever reason (i.e by manual database entry), this will mean they would be locked out, no?

I would understand this for the creation of passwords, however.

cassdegrees/templates/registration/login.html
j-selby: Na, something like `href="javascript:alert(...)"` is what I meant here

cassdegrees/templates/registration/password_change_done.html
j-selby: Typo - you called it `hompage`. not `homepage`.

cassdegrees/cassdegrees/settings.py
JackSekoranja: This is only for password creation. Insecure password still work normally

cassdegrees/templates/registration/password_change_done.html
JackSekoranja: Oh I see it now, thanks :')

j-selby (approved):
Looks good.

j-selby:
I can't see how passwords are actually stored in the database here, however. Are they stored as a model, and are passwords hashed/salted?

JackSekoranja:
Passwords are managed by Django directly. From my understanding they aren't stored in a database, and I can confirm they are all hashed and salted. Not even admins can get the full hash for passwords. More info here:
https://docs.djangoproject.com/en/2.2/topics/auth/passwords/

Issue #81, opened 2019-04-29
Modify requisites in course creation/editing to support complex requisites


JackSekoranja:
Currently course requisites and incompatibilities are created by selecting multiple courses from a list. However, not all requisites/incompatibilities are representable in this format, as some have logic such as "Completion of (A and (B or C))", and can even have additional criteria such as "Must get a permission code" or "Must be studying X". 

There are several approaches we could take to solve these issues:
1. Use program/subplan rules logic to represent this information (More complex than needed)
2. Have a simple course list that indicates a relationship with the course, and have a text section below describing the rules (Easy to do/understand but will lose some representation)
3. Use basic AND/OR logic to represent requisites, and an optional text input for other options
4. Mixture of above suggestions

Might want to discuss the best option to take with team (I personally think 3 is a good idea)

Outlier Course examples:
- Permission code & unit requirement: https://programsandcourses.anu.edu.au/2019/course/ASIA2107
- Must be studying X: https://programsandcourses.anu.edu.au/2019/course/BUSN7051
- Mixed rules: https://programsandcourses.anu.edu.au/2019/course/LAWS2201

Daniel-Jang:
Different Types of Requisites/Incompatibilities **(NOTE: requisites will be in an AND/OR logic format)**:
- "Inherent Requirements"??? - could be ignored, search for definition.
- X units of university courses: 
Eg. _24 units of university courses_
- X units of subject-specific courses: 
Eg. _12 units of COMP courses_
- X units of year-specific courses: 
Eg. _48 units of 2000-level courses_
- X units of subject and year specific courses: 
Eg. _12 units of COMP3000 courses_
- Specific courses: 
Eg. _must have completed LAWS1000_
- Concurrent enrolment: 
Eg. _You currently must be completing five 1000-level LAWS courses, You currently must be studying X_
- Must currently be studying a certain program:
Eg. _You must be studying a program which includes Bachelor of Laws_
- Must ask for a permission code
- Incompatible with X **(will only be represented using AND):**
Eg. _This course is incompatible with BUSN3051_

spencerj411:
done - see #156.

Issue #78, opened 2019-04-26
Add Course Rules


Daniel-Jang:
As a CASS Staff Member,
I want to be able to define the course rules with requisites and incompatible courses when creating a new course, so that I can save and record the requisites and incompatible courses for every course.

**Acceptance Criteria:**
- User can define requisites like such: "Must have done X or Y" or "Must have done A and B"
- User can define incompatible courses like such: "Incompatible with M and N"

spencerj411:
done - see #156.

PR #156, opened 2019-05-21
Course Requisites and Incompatibilities

spencerj411:
Aims to complete #81 and #78.
This pr contains the following additions/modification:

- added `rules` field to CourseModel
- added either/or rule to course
- added requisite/incompatibility section in the form
- added program field to requisite
- added incompatibility option
- added year-level field to subject-area units to satisfy all criteria (eg. 12 units of COMP 4000-level courses)

Note:
- The View feature will unfortunately have to be updated. #94 

**Whilst creating a course**
image

**Once it has been saved**
image


cassdegrees/templates/createcourse.html
```diff
@@ -42,31 +41,47 @@
...
-
-            <p>
-                <label for="text">Incompatible Course(s)</label>
-                <select class="text tfull" name="incompatibles[]" size="5" multiple>
-                    {% for course in courses %}
-                        <option value="{{ course.code }}">{{ course.code }}</option>
-                    {% endfor %}
-                </select>
-            </p>
+{#            <!-- TODO: Unhandled so far -->#}
```
j-selby: Wait, can this be removed now?

cassdegrees/templates/widgets/rulescripts.html
```diff
@@ -0,0 +1,202 @@
...
+                {{ component_names[details.type] }}
+            </p>
+            <input type="button" v-on:click="$emit('remove')" class="btn-uni-grad btn-snall no-left-margin" value="Remove" />
+        </header>
+        <div class="box-solid">
+            <div class="card-content" v-bind:is="'rule_' + details.type" v-bind:details="details"></div>
+        </div>
+    </div>
+</script>
+{% endverbatim %}
```
j-selby: Newline

j-selby (approved):
Nice work!

JackSekoranja (changes requested):
I did find a few things that we might want to change:
- 'Course' and 'Incompatibility' should work the same. Maybe remove the unit count from 'Course' and re-name it 'Course Requisite'
- It says "Students must be studying a program which includes:", but you can only select one option
- There is a unit count on custom-text items. We might not want this since it may be things like "Students need to get a permission code for this course"
- Units is also mentioned in other places (Subject-area, Level-specific), but it may not make sense in this context

Screen Shot 2019-05-22 at 2 00 11 pm

Otherwise it looks very good, well done :)

Daniel-Jang:
@JackSekoranja I had a look at your comment and the first 3 dot points I agree with. However, with your 4th point, there are cases where a course requisite would be like "... and 6 units of 1000 level MATH":
https://programsandcourses.anu.edu.au/2019/course/COMP2100

I will fix everything else, however, since you raised valid points about them.

Edit: Completed the first 3 suggestions, ready for review.

JackSekoranja (approved):
Looks good

PR #158, opened 2019-05-22
Vectorise icons, add icons to dropdown menu

j-selby:
Vectorised icons throughout the application to fix issues with blurriness.

Dropdown menu for user was a little hard to identify - added a dropdown icon to represent this functionality.

example-homepage


KavsHub (approved):
Looks good to me.
One thing that was raised by the UAT was to make the home button larger - which couldn't be done previously as it wasn't a vector image. May want to play with the sizing a little.

j-selby:
@KavsHub this is already larger than the previous version - I would wait to see how this one goes before we  go too far

JackSekoranja (changes requested):
Got a bug in mine where it looks like this on Chrome:
Screen Shot 2019-05-23 at 8 00 25 pm

Works fine on safari

j-selby:
@JackSekoranja did you refresh your CSS? Added a few rules in there for that

JackSekoranja:
Yeah, this was in incognito mode

j-selby:
Just loaded it up in Chrome and it worked fine for me there. The bar in the top left is just a border rule with some flavour, which would suggest that the css isn't loading correctly. Could you dump what Chrome sees in devtools for the "menu-divider" element?

JackSekoranja (approved):
Looks like it has fixed itself. It turns out you need to close and reopen incognito mode for the CSS cache to be deleted :/ . Happy for you to merge

PR #159, opened 2019-05-22
Added Cancel Buttons to Create/Edit pages

KavsHub:
As requested, there's now a cancel button on the create/edit pages.


JackSekoranja (changes requested):
Works well, but could you do the same for program/subplan/course deletion? After that I believe that will be everything we need

KavsHub:
> Works well, but could you do the same for program/subplan/course deletion? After that I believe that will be everything we need

I thought there was already yes - no buttons but apparently not anymore. I've updated them now

JackSekoranja (approved):

cassdegrees/static/js/main.js
```diff
@@ -10,4 +10,18 @@
...
+function goBack() {
+    window.location.href= "/";
+}
+
+/**
+ * Returns to the relevant list page
+ */
+function returnToList(pageName) {
+    window.location.href = "/list/?view="+pageName;
 }
```
j-selby: Newline

Issue #104, opened 2019-05-07
Prevent course name changing if a degree requires it


JackSekoranja:
We've done checks to prevent against users deleting courses if they appear in subplans and programs, but it is still possible to rename a course without these dependencies being checked. By doing this, users would be able to bypass the deletion checks and leave the system in an inconsistent state

spencerj411:
done - see #163.

PR #163, opened 2019-05-27
Prevent Course Code Change if Program/Subplan Requires It

spencerj411:
Aims to complete #104. How I've tackled this issue is I check if a course code is being used in any subplans/programs. If it is, I disable the Course Code input box and display an info message. If no subplan/program depends on it, you're able to edit the course code as normal.

**When you're editing a course that is needed by a subplan/program:**
image
**When you're editing a course that is _not_ needed by a subplan/program:**
image


JackSekoranja (approved):
Looks good and works well. We may want to do the same thing for course requisites but since we have overlooked this in the course deletion rules as well, it should be fine to add both in a separate PR

spencerj411:
sounds good. will merge now.

PR #164, opened 2019-05-28
Fix use of requests in view

j-selby:
Production server doesn't bind to a TCP address for HTTP requests. This PR fixes this by using proper Models.


JackSekoranja (approved):
Seems to work well. No noticeable negative side effects.

Issue #167, opened 2019-05-29
PDF and View on server resulting in error


joshcbee:
![image](https://user-images.githubusercontent.com/10462891/58535725-f4962600-8231-11e9-89aa-612ed45371e9.png)

![image](https://user-images.githubusercontent.com/10462891/58535754-feb82480-8231-11e9-91c1-f16a75830e04.png)

j-selby:
@JackSekoranja can this be closed then?

JackSekoranja:
While invalid data will still cause this issue, it is no longer possible to input invalid data in to the system from the staff user interface (PR #168).

PR #168, opened 2019-05-29
Added strict checks to ensure blank rules cannot be submitted

JackSekoranja:
This PR addresses issue #167. The original issue was that someone was allowed to enter a subplan with the id `-1` (i.e. They chose to add a subplan and didn't select anything). This PR adds proper verification, making it impossible to add a rule without entering in any information.


KavsHub (approved):
Good work on finding and fixing the bug. Looks good to me :)

PR #173, opened 2019-06-03
Implement either/or for PDF program templates

### LINKED ISSUE #95: Update Program/Subplan View Pages With New Rules
As new program and subplan rules are added to the system, the view pages will need to be updated to neatly display these.

j-selby:
Closes #95 (see comment in there).

This PR implements the either/or rule for PDF templates:

either-or-pdf

Rules for PDFs have been migrated to a recursive widget.

### DIFF
--- a/cassdegrees/templates/pdf_base.html
+++ b/cassdegrees/templates/pdf_base.html
@@ -79,6 +79,10 @@
         .top-margin {
             padding-top: 0.1cm;
         }
+
+        .try-complete-box {
+            page-break-inside: avoid;
+        }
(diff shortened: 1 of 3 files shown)


JackSekoranja (approved):
For me the rule can skip across to the next column and it looks kind of awkward. Would be great to fix but it does work

PDF

j-selby:
@JackSekoranja I've added some hinting in there - does this display nicely now?

JackSekoranja:
Yep, that's fixed it

PR #177, opened 2019-08-06
UAT insert link to conflicting programs, subplans and courses when a …

joshcbee:
…uniqueness constraint is violated.

The error will only trigger when valid entries have been provided for the constrained fields, as field-level validation is run prior to form level.

The relevant constraints to test are:
- program - unique_together = (("code", "year"), ("name", "year"))
- subplan - unique_together = (("code", "year"), ("year", "name", "planType"),)
- course - unique_together = (("code", "year"),)

image


cassdegrees/ui/forms.py
```diff
@@ -89,6 +108,45 @@ def clean_units(self):
...
         return data
 
+    # Override clean to return links to existing content if unique_together constraint fails
+    # https://stackoverflow.com/questions/4659360/get-django-object-id-based-on-model-attribute
+    def clean(self):
+        cleaned_data = super().clean()
+
+        # check assignment as keys may not exist in cleaned dictionary if field level validation has failed
+        try:
+            draft_code = cleaned_data['code']
```
j-selby: Is there no way to hook into the pre-existing error checking here? No biggie if not.

j-selby (approved):
Looks good otherwise!

cassdegrees/ui/forms.py
joshcbee: We're hooking into the pre-existing clean method. But the pre-existing validate_unique() method is a nightmare and goes several layers down (multiple overrides) and from what I can see has no easy place to hook in to actually return a conflicting id.

This method has the benefit that the natural errors raised by validate_unique will specify the conflicting fields with the simple override we've already put in the NON_FIELD_ERRORS, so we can make some generic 'get conflicting record link' methods.

JackSekoranja (approved):
Looks good and works well, happy for you to commit

Issue #116, opened 2019-05-09
Create "Save" and "Save and Exit" Buttons for Editing


aidengeorge97:
UAT revealed that some users would like to be able to Save an item they are editing before leaving the editing page. This reassures them that the data they have entered has not changed upon leaving the page, and that they submitted correctly. 

Editing pages should implement a "Save" button, that saves the item to the database without leaving the page, and a "Save and Exit" button that behaves as the current button does.

Daniel-Jang:
Pull Request: https://github.com/cass-degrees/CASS-Degrees-Code/pull/178

PR #178, opened 2019-08-06
Add "Save" and "Save and Exit" Buttons for Editing

Daniel-Jang:
Original UAT Issue: https://github.com/cass-degrees/CASS-Degrees-Code/issues/116

I have added the 'Save' button that saves the details inserted by the user, without redirecting them back to the list page.

image

### DIFF
--- a/cassdegrees/templates/createcourse.html
+++ b/cassdegrees/templates/createcourse.html
@@ -24,6 +24,8 @@
     <form class="anuform" id="mainForm" action="" method="post" onsubmit="return false;">
         {% csrf_token %}
 
+        <input type="hidden" id="redirect" name="redirect">
+
         <fieldset>
             <legend>Course details</legend>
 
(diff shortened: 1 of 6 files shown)


cassdegrees/templates/createcourse.html
```diff
@@ -56,15 +61,21 @@
...
     </p>
 
     <script>
-        function submit_form(form_action) {
+        function submit_form(form_action, redirect) {
             if (handleRules()) {
                 document.getElementById('mainForm').action.value = form_action;
+                document.getElementById('redirect').value = redirect;
                 document.getElementById('mainForm').submit();
+                console.log(document);
```
j-selby: Unneeded debugging (document object is going to be huge!)

j-selby (changes requested):
Looking good!

JackSekoranja (approved):
Looks good, although I agree with James' suggested changes. I did notice that there was no 'Cancel' button when editing courses, despite there being one for Programs and Subplans. I know that isn't part of this pull request, but it's similar enough to add in this pull request since we'll need to fix it anyway.

Daniel-Jang:
I've done what James and Jack recommended, and now is ready to merge.

j-selby (approved):
Looks good!

PR #211, opened 2019-08-10
Create student page and reorganise directories

Daniel-Jang:
Original Issue: https://github.com/cass-degrees/CASS-Degrees-Code/issues/179

This was done so that the main page (127.0.0.1:8000) directs the users to the student optimised home page, and everything we have done in semester 1 has been shifted by '/admin/' prefix. 

The exception is the old index page, which is now '/admin/home/' since '/admin/' leads to the Django admin page.

All the relevant files (HTML, JS, Django) has been modified to ensure that the hyperlinks to other pages are valid.


cassdegrees/static/js/main.js
```diff
@@ -16,12 +16,12 @@ function checkKeys(event) {
  * Makes the web page go back to the homepage
  */
 function goBack() {
-    window.location.href= "/";
+    window.location.href= "/admin/home/";
```
j-selby: Can we use relative links here maybe to prevent this from being a problem in the future?

cassdegrees/templates/bulkupload.html
```diff
@@ -6,6 +6,7 @@
 {% block subtitle %}Bulk Data Upload{% endblock %}
 
 {% block breadcrumb-trail %}
+    {% breadcrumb '/admin/home' 'Admin Home'%}
```
j-selby, line 9: Quick consistency bit: space after `home'`

cassdegrees/templates/bulkupload.html
```diff
@@ -25,7 +26,7 @@
 
     {% for type in upload_type %}
         <div id="{{ type }}Page" hidden>
-            <form class="anuform" action="/bulk_upload/?type={{ type }}" method="post" enctype="multipart/form-data">
+            <form class="anuform" action="/admin/bulk_upload/?type={{ type }}" method="post" enctype="multipart/form-data">
```
j-selby: Can this be relative? (i.e `action="?type={{ type }}"`)?

cassdegrees/templates/list.html
```diff
@@ -20,7 +21,7 @@
         <p class="msg-error">{{ render.error|linebreaksbr }}</p>
     {% endif %}
 
-    <form class="box bdr-solid bdr-uni bg-uni50 anuform custom_search_bar" action="/list/?action=Search" method="get">
+    <form class="box bdr-solid bdr-uni bg-uni50 anuform custom_search_bar" action="/admin/list/?action=Search" method="get">
```
j-selby: Ditto as above.

cassdegrees/templates/student_index.html
```diff
@@ -0,0 +1,15 @@
...
+
+{% block subtitle %}Student Home{% endblock %}
+
+{% block breadcrumb-trail %}
+    {% finalcrumb 'Student Home' %}
+{% endblock %}
+
+{% block content %}
+
+{% endblock %}
```
j-selby: Missing newline.

j-selby (changes requested):
Looks pretty good. PEP also has a few issues above if you haven't seen them.

j-selby:
As a quick nitpick, can we also migrate admin views somewhere else?

cassdegrees/static/js/main.js
Daniel-Jang: I have done it slightly differently, take a look when I push up my changes.

cassdegrees/templates/bulkupload.html
Daniel-Jang: There is one space after home, should there be two? I am not sure

cassdegrees/templates/bulkupload.html
Daniel-Jang: Yep, tried that and it works!

cassdegrees/templates/student_index.html
Daniel-Jang: Ahhh, my bad once again hahaha

Daniel-Jang:
> As a quick nitpick, can we also migrate admin views somewhere else?

Still technically UI, but do you wish to move it into a subfolder called 'admin' or something similar? I think it is not a problem right now as-is, but we can bring this up in the next group discussion.

j-selby:
Well, right now we have templates + views + etc all sharing space - could do with some cleanup to help see what is what.

cassdegrees/templates/bulkupload.html
j-selby: No, at the end of 'Admin home', sorry

j-selby (approved):
With my other nitpicks aside, looks good!

aidengeorge97 (changes requested):
This looks good, but one issue I noticed is that some links still go back to 127.0.0.1:8000 rather than the new admin/home page. For example, if you go to create a course and then cancel, you will be sent to the student home page rather than the admin home page. Cancelling from the edit program screen results in an error.

Daniel-Jang:
@aidengeorge97 I will investigate this, thank you

Daniel-Jang:
> This looks good, but one issue I noticed is that some links still go back to 127.0.0.1:8000 rather than the new admin/home page. For example, if you go to create a course and then cancel, you will be sent to the student home page rather than the admin home page. Cancelling from the edit program screen results in an error.

@aidengeorge97 Have you tried clearing cache? Try hitting CTRL + SHIFT + R on the course creation page, then click cancel because I end up at /admin/home when I follow your steps.

aidengeorge97 (approved):
Ahh that seems to have fixed it. Sorry for that. I'm happy to approve this now.

PR #214, opened 2019-08-12
Disabled autocomplete on form inputs

### LINKED ISSUE #161: Disable HTML Form Input Autocomplete and Autofill
Add autocomplete="off" onto <form> element;

Add hidden <input> with autocomplete="false" as a first children element of the form.

aidengeorge97:
Closes #161 

This PR adds a tag to the template creation forms which prevents web browsers from showing autocomplete suggestions. UAT showed that autocomplete could be misleading for some users, and would not benefit the user workflow.

I had some trouble testing this as I do not experience any autocomplete suggestions on my environment (Chrome and Microsoft Edge). The change is small and should work in theory however. It would be good for people who experience autocomplete suggestions currently to validate whether this PR disables them.

### DIFF
--- a/cassdegrees/templates/createcourse.html
+++ b/cassdegrees/templates/createcourse.html
@@ -24,7 +24,7 @@
          <p class="msg-error">{{ error }}</p>
     {% endfor %}
 
-    <form class="anuform" id="mainForm" action="" method="post" onsubmit="return false;">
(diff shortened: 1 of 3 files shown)


Daniel-Jang (approved):
Nothing looks wrong on my end.

j-selby (approved):
Looks good as well

PR #215, opened 2019-08-14
Corrected to check whether there is an instance ID that can be used t…

joshcbee:
…o determine whether conflicts are due to updating the existing record


j-selby (approved):
Tested and looks good.

PR #226, opened 2019-08-15
Updated SETUP.md to get the bullet points to render correctly

Daniel-Jang:
The bullet points weren't rendering properly so I updated it to make it display correctly like below:
Before:
image

After:
image


Daniel-Jang:
There isn't any real content or feature that was added, rather it was more like a very simple bugfix. So I will merge this to develop now.

Issue #216, opened 2019-08-14
Can't tab to next field


joshcbee:
Can't tab to next field from Year and Units fields

KavsHub:
Bug fixed and merged in PR #231

PR #231, opened 2019-08-21
Fixed UAT Bug #216 Unable tab number fields

KavsHub:
I figured out why we couldn't tab number fields. Issue (#216)
This was because our Javascript function to stop 'e' and '+' '-' used in number fields was also rejecting the tab keycode.

I've simply added it as an exception in out checkKeys function.


JackSekoranja (approved):
Works perfectly, well done. Happy for you to merge

Daniel-Jang (approved):
Works great!

Issue #202, opened 2019-08-07
Save Plan (8)


spencerj411:
As a _Student_,
I want to be able to save my plan,
So that I can come back and continue working on it later.

**Acceptance Criteria:**

- After saving a plan and exiting the browser, it is possible to return to the saved plan

**User Story Points:** 8

JackSekoranja:
Closed with PR #232

Issue #187, opened 2019-08-07
Add Timestamp for Plan (2)


spencerj411:
As a _Student_,
I want my plan to contain a timestamp that displays the time and date it was last updated,
So that I can tell how recent the plan is.

**Acceptance Criteria:**

- Plan displays the time and date it was last updated
- Plan updates its timestamp when it is edited.
- Plan does not update its timestamp when it is loaded but no changes are made.

**User Story Points:** 2

JackSekoranja:
Closed with PR #232

Issue #184, opened 2019-08-07
Plan Deletion (2)


spencerj411:
As a _Student_,
I want to be able to delete plans I no longer need,
So that I can restrict the number of existing plans to contain only relevant ones.

**Acceptance Criteria:**

- No information about the deleted plan is retained in any manner

**User Story Points:** 2

JackSekoranja:
Closed with PR #232

Issue #183, opened 2019-08-07
Load Previously Worked On Plan (5)


spencerj411:
As a _Student_,
I want to be able to load a plan I had started previously,
So that I can continue working on it.

**Acceptance Criteria:**

- Loading up a plan doesn’t affect any other plan

- The loaded plan retains the same data it had when saved

**User Story Points:** 5

JackSekoranja:
Closed with PR #232

PR #232, opened 2019-08-21
Feature plan save and load

JackSekoranja:
This PR implements Issues #202, #183, #187, and #184. The primary new feature is plan saving and loading using the main menu system. This change also modifies the `student_edit.html` file by using a form and submit button rather than a link. 

The saving and loading uses `zlib` compression library, as it was experimentally found to give the best compression results for longer json strings. `bz2` and `lzma` were also tested.

**Adding New Plan Elements**
To add new details to the plan, one only needs to add an input element as with the following attributes inside the `form` element:
- `name="[Field Name]"`: The name that should be used to refer to the object
- `value="{{ plan.[Field Name] }}"`: Using that object to load existing plans

After doing this, the plan should automatically take the value inputted by the user when saving, and restore it when loading

Screen Shot 2019-08-21 at 10 47 08 pm

### DIFF
--- a/cassdegrees/cassdegrees/settings.py
+++ b/cassdegrees/cassdegrees/settings.py
@@ -120,7 +120,7 @@ AUTH_PASSWORD_VALIDATORS = [
 
 LANGUAGE_CODE = 'en-us'
 
-TIME_ZONE = 'UTC'
+TIME_ZONE = 'Australia/Sydney'
 
 USE_I18N = True
 
--- a/cassdegrees/static/css/style.css
+++ b/cassdegrees/static/css/style.css
@@ -53,6 +53,21 @@
     width: 100%;
 }
 
+.large-menu-item {
+    width:50%;
+}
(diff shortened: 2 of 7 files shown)


aidengeorge97 (changes requested):
Saving a plan with no name entered returns a "sucessfully saved" message, but the plan is not actually saved. It should return an error message if you try to save without naming the plan.

Daniel-Jang:
The last modified timestamp does not show the exact time. It would be nice to see the hour and minute it was saved, as a user can work on the same plan multiple times a day. (Also it is part of the acceptance criteria for #187 ).

Also, there should be a confirmation prompt when you click on the 'Remove' button, since it is very easy to accidently click on it.

Otherwise, very good!

Daniel-Jang (approved):

aidengeorge97 (approved):
With the new changes, this is ready for merge.

Issue #217, opened 2019-08-14
Make Bulk Upload Compatible with CASS' .xlsx Format


spencerj411:
**(FEEDBACK FROM CASS' USER TESTING)**

> "_Is it possible to do a mass upload of courses from the 3-year teaching plan (stored, I think, as a spreadsheet)?_" -Deborah Veness

**Important Links:**
CASS’ UAT: https://docs.google.com/document/d/1YgSb5DiK-O8mrNe_oHAtijXuIVcSzRwb/edit
Excel Spreadsheet Format: https://drive.google.com/open?id=1EdQHhrNSd2QhO4c2A_su3x3K7hHQKQoI

Daniel-Jang:
I've just noticed the excel spreadsheet format after working on this ticket for a bit, will need to recode again to support that format but everything else works fine. Was about to submit a PR but will have to wait a bit unfortunately.

j-selby:
@Daniel-Jang would it be easier to get them to export as a .csv from Excel? Especially if they shift their workflow to e.g. Google Docs or something in the future?

Daniel-Jang:
> @Daniel-Jang would it be easier to get them to export as a .csv from Excel? Especially if they shift their workflow to e.g. Google Docs or something in the future?

Whoops, sorry I didn't see this earlier :/ 
Thankfully I got it working with excel though, and they would not need to do anything on their end in terms of formatting the file.

PR #233, opened 2019-08-21
Make Bulk Upload Compatible with CASS' .xlsx Format

Daniel-Jang:
Original Issue: https://github.com/cass-degrees/CASS-Degrees-Code/issues/217

Added support for Excel sheets, either in the custom format that the bulk uploader could already support, or in the format CASS uses.

To test this branch out, try the following file: https://drive.google.com/open?id=1EdQHhrNSd2QhO4c2A_su3x3K7hHQKQoI

Note that not all will be successfully added, but most will be.

### DIFF
--- a/cassdegrees/ui/views/bulk_data_upload.py
+++ b/cassdegrees/ui/views/bulk_data_upload.py
@@ -90,15 +90,17 @@ def bulk_data_upload(request):
                     # where the file reader determines the year ranges and column positions.
                     if not year_range_found or not columns_set:
                         for cell in row:
-                            # Split the string in the first line of the excel sheet to get start and end year
+                            # Split the string in the first line of the excel sheet to get start and end years.
                             if not year_range_found:
                                 if cell.value is not None:
                                     stripped_title = cell.value.split()
 
-                                    start_year = int(stripped_title[-3])
-                                    end_year = int(stripped_title[-1])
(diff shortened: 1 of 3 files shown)


JackSekoranja (changes requested):
Functionally it works well, but if an incorrect excel file is chosen the website will crash, so it would be good to make it more robust regarding incorrect xlsx files. I also think it is unnecessary to display courses that didn't upload due to already being in the system, especially given the large number of courses that failed.

This also may be too difficult (so take this feedback as optional) but since there are so many courses on the error page, it might be useful to be able to hide some of them somehow.

JackSekoranja (approved):
Works, would like you to prevent duplicate courses from printing but otherwise good

cassdegrees/ui/views/bulk_data_upload.py
```diff
@@ -19,10 +20,15 @@
 # ARTI-SPEC%2016%Artificial Intelligence%24%SPEC
 # ...
 
+# 20/08/19 changes: Support added for excel sheets in the same rules as above (no % needed, but put each word between
+# the % sign in a new cell in that row.
+
+# 21/08/19 changes: Support added for CASS teaching plan excel file.
```
j-selby: Shouldn't this be left to commits?

j-selby (approved):
Looks good!

Issue #208, opened 2019-08-07
Openable/Editable Plan via Email (8)


spencerj411:
As a _Student/CASS Staff Member_,
I want to be able to send my plan in an email,
So that I can get CASS Staff/Students to view my changes in the plan

**Acceptance Criteria:**

- The plan must be able to be opened and edited by the party receiving the email

**User Story Points:** 8

JackSekoranja:
Closed with PR #242

PR #242, opened 2019-08-24
Feature email send and load

### LINKED ISSUE #186: View & Edit Plan from Email (8)
As a _Student_,
I want to be able to view and edit a plan that I have received/sent in an email,
So that I can modify that plan on other computers and review plans I have sent to CASS Staff.

**Acceptance Criteria:**

- Students should be able to open plans in the same way that a CASS Staff member can.
- Plans that students open must be fully editable, and function as if the plan had been opened from the selection menu.
- Writing to the plan will leave the Staff comments intact (if applicable), but will remove the ‘prepared by’ date, indicating an unfinalised plan.

**User Story Points:** 8

JackSekoranja:
This PR closes #186 and #208.

A new "Export" option has been added to the plan saving menu. Upon clicking, a popup appears with a link that can be copy-pasted to reload the plan.

Screen Shot 2019-08-24 at 12 54 30 pm

### DIFF
--- a/cassdegrees/templates/student_edit.html
+++ b/cassdegrees/templates/student_edit.html
@@ -42,12 +42,6 @@
                 <input type="button" onclick="prepareSubmit('approve')" class="btn-uni-grad btn-small" value="Save & Approve...">
             {% endif %}
         </div>
-
-        {% if render.popup %}
-            <p class="anuform">
-                {{ render.popup |safe }}
-            </p>
-        {% endif %}
     </form>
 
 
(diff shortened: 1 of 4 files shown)


Daniel-Jang (changes requested):
When you copy and paste the link in a different browser, saving the plan will result in this error:

image

I don't think changing browsers causes this, but I used a browser that would not have the same caches as my main brower.

KavsHub (approved):
Works for me now; Tested on different browsers too.

cassdegrees/templates/student_edit.html
```diff
@@ -101,6 +104,44 @@ <h3>Courses</h3>
...
+                <div id="modal_content"></div>
+            </div>
+            <footer class="box-solid">
+                <input type="button" id="copy_to_clipboard" class="btn-uni-grad btn-small" value="Copy to Clipboard" />
+                <input type="button" id="close_modal" class="right btn-uni-grad btn-small" value="OK" />
+            </footer>
+        </div>
+        {# Script to allow interactivity in the menu #}
+        <script type="application/javascript">
+            var plan_link = document.getElementById("plan_link");
```
j-selby, line 129: Can this be extracted to a js script? In student_edit maybe?

j-selby (approved):
This PR seems to have broken previously saved plan names (related to the issues in #243?), but looks good otherwise!

cassdegrees/ui/views/student.py
```diff
@@ -49,6 +49,13 @@ def load_messages(cookies):
             del cookies['message']
         except KeyError:
             pass
+    popup = cookies.get('popup', None)
```
j-selby, line 52: Do we need to use cookies for this one? This seems a bit round-about. It is possible to do this clientside?

cassdegrees/templates/student_edit.html
JackSekoranja: I can move this, but if I do it may confuse future readers as there is no indication that the three elements are using javascript. Should I still extract the code?

cassdegrees/ui/views/student.py
JackSekoranja: Cookies aren't needed, although I don't know any other way to do this without either using cookies or displaying it in the URL parameters, and I feel this way is a bit cleaner on the user-end

Daniel-Jang (approved):

cassdegrees/templates/student_edit.html
j-selby: Preferably, yeah. There are cross-site scripting issues down the line for inline Javascript - not relevant now, and there are a couple of other similar instances in the cost base, but something to keep in mind. Ideally the relevant elements would be commented.

JackSekoranja: Okay, I added a fix to this in PR #243

PR #243, opened 2019-08-24
Add duplication functionality

### LINKED ISSUE #185: Plan Duplication (3)
As a _Student_,
I want to be able to duplicate an existing plan,
So that I can make changes and compare different possibilities for a plan without editing the original.

**Acceptance Criteria:**

- After duplicating a plan, both the original plan and the copy exist and can be modified independently. 

- It is possible to easily identify which plan is which.

**User Story Points:** 3

JackSekoranja:
This PR closes #185 

The "duplicate" button in the main menu is now functional, and will create a copy of the plan

### DIFF
--- a/cassdegrees/ui/views/student.py
+++ b/cassdegrees/ui/views/student.py
@@ -88,6 +88,7 @@ def student_delete(request):
 # Creation page. Also sends program metadata.
 def student_create(request):
     id_to_view = request.GET.get('id', None)
+    plan_to_duplicate = request.GET.get('plan', None)
 
     # Create a plan if an ID is specified
     if id_to_view:
@@ -100,6 +101,23 @@ def student_create(request):
         response = redirect(student_edit)
         response['Location'] += '?plan='
         return response
+    # If a plan name is given, copy that plan
+    elif plan_to_duplicate:
+        plan = decompress(request.session.get('plan:'+plan_to_duplicate, None))
+        print(plan)
+        if plan:
+            new_plan_name = plan['name'] + ' (Copy)'
+            plan['name'] = new_plan_name
+            plan['date'] = timezone.localtime().strftime('%d/%m/%Y %H:%M')
+            request.session['plan:' + new_plan_name] = compress(plan)
+
+            # Redirect to the student edit page and add the '?plan=' url parameter
+            response = redirect(student_edit)
+            response['Location'] += '?plan='+new_plan_name
+            return response
+        else:
+            request.session['error_message'] = "Invalid plan name given"
+            redirect(student_index)
     # Render the creation homepage if an ID is not specified
     else:
         return render(request, 'student_create.html', context={'programs': ProgramModel.objects.filter(publish=True)})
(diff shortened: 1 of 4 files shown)


Daniel-Jang (approved):
Works well for me, happy for it to be merged.

KavsHub:
Hey, I'm getting an error when I try to duplicate my plan.
image
Am I doing something wrong?

JackSekoranja:
Ah yes, I believe this is happening because you are duplicating a old plan, which saved different data. Try creating a new plan and duplicating that

j-selby:
I'm getting the same error with a new plan.

Steps to reproduce:
- Create a new plan based on a pre-existing program
- Add a course
- Save
- Head to main menu
- Press duplicate, same error as above

cassdegrees/ui/views/student.py
```diff
@@ -100,6 +101,23 @@ def student_create(request):
         response = redirect(student_edit)
         response['Location'] += '?plan='
         return response
+    # If a plan name is given, copy that plan
+    elif plan_to_duplicate:
+        plan = decompress(request.session.get('plan:' + plan_to_duplicate, None))
+        print(plan)
```
j-selby: Make sure you aren't logging this when you merge

cassdegrees/ui/views/student.py
```diff
@@ -100,6 +101,23 @@ def student_create(request):
...
+        if plan:
+            new_plan_name = plan['name'] + ' (Copy)'
+            plan['name'] = new_plan_name
+            plan['date'] = timezone.localtime().strftime('%d/%m/%Y %H:%M')
```
j-selby, line 122: Are we still using localtime? Did we come to a consensus on this?

cassdegrees/ui/views/student.py
```diff
@@ -100,6 +101,23 @@ def student_create(request):
...
+            request.session['plan:' + new_plan_name] = compress(plan)
+
+            # Redirect to the student edit page and add the '?plan=' url parameter
+            response = redirect(student_edit)
+            response['Location'] += '?plan=' + new_plan_name
```
j-selby: Can you do this in one call in redirect (e.g. `redirect('?plan=' + new_plan_name)`? Re configuring HTTP headers could potentially be error-prone.

j-selby (changes requested):
Other than the error above and my few comments below, looks good!

cassdegrees/ui/views/student.py
JackSekoranja: Yeah I just noticed this, removed.

cassdegrees/ui/views/student.py
JackSekoranja: We needed to use it as otherwise we would be giving the wrong time to the users. This instance is a purely front-facing feature so using localtime isn't a huge issue

cassdegrees/ui/views/student.py
JackSekoranja: Yes I can, but it would need to be `redirect('/edit/?plan='+new_plan_name)`. I figured doing it this way requires fewer rewrites if we want to change the address later, although I might just do it the way you are suggesting as that's used in other areas of the code.

Issue #252, opened 2019-08-25
Either/or rule: No intuitive way to delete a group


j-selby:
The either/or rule allows for the creation of groups, but empty groups are left as is with no intuitive way to delete them.

Screenshot:

![empty-group](https://user-images.githubusercontent.com/1404334/63648436-4a8a4500-c71f-11e9-9704-7218f67c2097.png)

PR #273, opened 2019-08-27
Added the ability for users to remove or groups in 'either or ' rules.

Daniel-Jang:
Original Issue: #252 

The users now have the ability to remove groups in 'either or' rules section.
image


j-selby (approved):
Works well - looks good.

PR #274, opened 2019-08-27
Add ability to remove first course in a rule if >1 course is defined

### LINKED ISSUE #264: Ability to Remove First Course from a Course Rule
When I have multiple courses in a course rule, I would like to be able to remove the first course directly, rather than having to switch courses with a removable course and removing that instead. 

image

j-selby:
Closes #264.

This PR is a simple fix to allow for the first course in a course selection to be removed if there is more than one course in that list.

Screenshots:

fixed-courses-1
fixed-courses-2


Daniel-Jang (approved):
Works well for me. You can merge it in.

Issue #277, opened 2019-08-27
Limit the size of the list of items being deleted in the Deletions page


Daniel-Jang:
Currently, there is no limit on how many programs, subplans or programs are displayed at once. This means that hypothetically, if a 1000 courses were selected to be deleted, the message would be 1000 courses long.

There should be a limit on how many items are displayed at once (5 perhaps)

Daniel-Jang:
We decided we will not go through with this as the user should be able to see the full list of items they are about to delete.

PR #279, opened 2019-08-28
Limit deletion page message size to 5 items

Daniel-Jang:
Original Issue: #277 

The deletion page now only displays at most 5 courses to avoid an incredibly long page like as such:
image

### DIFF
--- a/cassdegrees/ui/views/courses.py
+++ b/cassdegrees/ui/views/courses.py
@@ -121,7 +121,7 @@ def delete_course(request):
         # Only insert the "... and x more courses" if length is more than show_count.
         if num_instances > show_count:
             if num_instances == show_count + 1:
-                display_instances.append("... and {} more course".format(len(instances) - show_count))
+                display_instances.append("... and 1 more course")
             else:
                 display_instances.append("... and {} more courses".format(len(instances) - show_count))
 
--- a/cassdegrees/ui/views/courses.py
+++ b/cassdegrees/ui/views/courses.py
@@ -114,8 +114,20 @@ def delete_course(request):
 
         return redirect(list_course_url + '&msg=Successfully Deleted Course(s)!')
     else:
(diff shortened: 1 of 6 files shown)


cassdegrees/ui/views/courses.py
```diff
@@ -114,8 +114,20 @@ def delete_course(request):
...
         return redirect(list_course_url + '&msg=Successfully Deleted Course(s)!')
     else:
+        # To avoid the delete confirmation page from being too long, only show the first 5 courses.
+        show_count = 5
+        num_instances = len(instances)
+        display_instances = instances[:show_count]
+        # Only insert the "... and x more courses" if length is more than show_count.
+        if num_instances > show_count:
+            if num_instances == show_count + 1:
+                display_instances.append("... and {} more course".format(len(instances) - show_count))
```
j-selby: `and 1 more course` would also work here - you already imply that constraint

cassdegrees/ui/views/courses.py
```diff
@@ -114,8 +114,20 @@ def delete_course(request):
...
+            else:
+                display_instances.append("... and {} more courses".format(len(instances) - show_count))
+
         return render(request, 'deletecourses.html', context={
-            "instances": instances
+            "instances": instances,
```
j-selby, line 129: Do you still need instances?

j-selby (approved):
Looks good!

Daniel-Jang:
Unfortunately, actually trying to delete this code will crash the server since it is trying to delete a course named "...and xx other courses".

Also, had a conversation with Jack and he suggested that the user should be able to see the full list since they shouldn't really be deleting courses anyway, so I will put this PR on hold for now.

JackSekoranja:
I'm going to close this for now as at present there seems to be no intention of progressing further with it. If in future we decide to keep this functionality we can re-open and merge it

PR #229, opened 2019-08-21
Extra Subplan Rules

spencerj411:
Closes #218 

There is now a new dropdown menu where the user can select min/max/exact for the rule. I had to use instyle css as a temp fix to overwrite anu styling (for some reason, could not do it through style.css).

**Adding a Course List Rule**
image

**How It Is Saved**
**NOTE: `different list types = {'max', 'exact', 'min'}`**
`{
   "id":17,
   "code":"RULE-MAJ",
   "year":2020,
   "name":"Testing Extra Rules",
   "units":48,
   "planType":"MAJ",
   "rules":[
      {
         "type":"course",
         "codes":[
            "ARTH1007",
            "ARTH1000",
            "LMAO3000"
         ],
         "list_type":"max",
         "unit_count":"12"
      }
   ],
   "publish":true
}`

PR #280, opened 2019-08-28
Added Min/Max/Exact units from a Course List Rules to Program

### LINKED ISSUE #255: Create/Edit Program Template Missing Course Rules
Same with #229, the Course (List) Rule needs extra rules. Course Rule needs:
Min/Exact/Max number of Units from a Course List.

spencerj411:
closes #255.

Same solution as on #229.

image

Edit:
added an error check for when nothing has been selected for the dropdown.

### DIFF
--- a/cassdegrees/static/css/style.css
+++ b/cassdegrees/static/css/style.css
@@ -28,10 +28,6 @@
     margin: 12px;
 }
 
-.course_list_types {
-    margin-left: 0;
-}
-
 .comments {
     resize: none;
     height: 5em; /* 1em height = 1 row */
--- a/cassdegrees/static/css/style.css
+++ b/cassdegrees/static/css/style.css
@@ -28,6 +28,10 @@
     margin: 12px;
 }
 
+.course_list_types {
+    margin-left: 0;
+}
+
 .comments {
     resize: none;
     height: 5em; /* 1em height = 1 row */
--- a/cassdegrees/templates/widgets/rulescripts.html
+++ b/cassdegrees/templates/widgets/rulescripts.html
@@ -82,7 +82,8 @@
             <option v-for="(msg, type) in list_types" v-bind:value="type">{{ msg }}</option>
         </select>
         
-        <input class="text course_list_types" onkeydown="javascript: return checkKeys(event)" type="number" v-on:change="check_options" v-model="details.unit_count" min="0" step="6" max="1000" required> units
+<!-- {#       instyle css only temporary #} -->
+        <input style="margin-left: 0;" class="text" onkeydown="javascript: return checkKeys(event)" type="number" v-on:change="check_options" v-model="details.unit_count" min="0" step="6" max="1000" required> units
 
         <div class="msg-error" v-if="invalid_units">Unit count must be > 0!</div>
         <div class="msg-error" v-if="invalid_units_step">Unit count should be divisible by 6!</div>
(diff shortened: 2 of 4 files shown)


cassdegrees/templates/widgets/rulescripts.html
```diff
@@ -76,10 +76,15 @@
...
             </label>
-            <input class="text" onkeydown="javascript: return checkKeys(event)" type="number" v-on:change="check_options" v-model="details.unit_count" min="0" step="6" max="1000" aria-required="true" required>
-            units
         </p>
 
+         <select v-model="details.list_type" required>
+            <option v-for="(msg, type) in list_types" v-bind:value="type">{{ msg }}</option>
+        </select>
+
+<!-- {#       instyle css only temporary #} -->
```
j-selby, line 85: Don't merge this one if it can be avoided - move this to a css class if needed.

cassdegrees/templates/widgets/rulescripts.html
```diff
@@ -76,10 +76,15 @@
             <label>
                 Students must select...
             </label>
-            <input class="text" onkeydown="javascript: return checkKeys(event)" type="number" v-on:change="check_options" v-model="details.unit_count" min="0" step="6" max="1000" aria-required="true" required>
-            units
         </p>
 
+         <select v-model="details.list_type" required>
```
j-selby: There doesn't appear to be error checking for this one - it can be saved with an empty value.

j-selby (changes requested):
Few issues, but looks good. Note that this is likely going to break pre-existing rules, so planning a migration strategy might be useful here.

cassdegrees/templates/widgets/rulescripts.html
spencerj411: fixed with latest batch of commits

cassdegrees/templates/widgets/rulescripts.html
spencerj411: if I move it to style.css, it can't overwrite anu styling (`.anuform fieldset fieldset input` adds padding to the units input box). So I had to settle with the inline styling for the meantime. @JackSekoranja is aware of this decision of mine. But I agree, it 's not ideal.

Daniel-Jang (approved):
Works well, happy for you to merge it in once James' suggestions are addressed.

cassdegrees/static/css/style.css
```diff
@@ -28,6 +28,10 @@
     margin: 12px;
 }
 
+.course_list_types {
```
j-selby: `-` in css for consistency

j-selby (approved):
Looks good now otherwise!

cassdegrees/templates/widgets/rulescripts.html
```diff
@@ -76,10 +76,15 @@
...
-            <input class="text" onkeydown="javascript: return checkKeys(event)" type="number" v-on:change="check_options" v-model="details.unit_count" min="0" step="6" max="1000" aria-required="true" required>
-            units
         </p>
 
+        <select v-model="details.list_type" v-on:change="check_options" required>
+            <option v-for="(msg, type) in list_types" v-bind:value="type">{{ msg }}</option>
+        </select>
+
+<!-- {#       instyle css only temporary #} -->
+        <input style="margin-left: 0;" class="text" onkeydown="javascript: return checkKeys(event)" type="number" v-on:change="check_options" v-model="details.unit_count" min="0" step="6" max="1000" required> units
```
j-selby, line 86: Sorry, didn't realise you had merged this yet - you can use !important to override css - e.g. `padding-left: 0 !important`, so you can still use classes for this one.

Issue #199, opened 2019-08-07
Add Subplan to Program Plan (8)


spencerj411:
As a _Student_,
I want to be able to add subplans to my program plan,
So that I can indicate my intended/actual completion of that subplan

**Acceptance Criteria:**

- There exists a way in which to indicate what subplans are/will be completed by the program plan
- Adding a subplan to the program correctly adds the requirements of that subplan to the program plan

**User Story Points:** 8

JackSekoranja:
Closed with PR #298

PR #256, opened 2019-08-26
Feature - Add Subplan to Program Plan

aidengeorge97:
Addresses #199.

This PR will allow students to nominate which subplan they wish to complete, given the subplan rules of their program. 

image

Given the acceptance criteria of #199, this does not close the issue as it does not "add the requirements of that subplan to the program plan". It does complete the first criterion however, and may be appropriate for this prototype.

This feature was implemented before #235 was pulled. I believe there is room for cleaner integration of the two features in both UI and code, but in its current state this feature is functional.

PR #298, opened 2019-09-02
Feature integrated subplan

JackSekoranja:
Closes issue #199

This PR integrates Aiden's previous PR (#256) with the course list page. Furthermore, it fixes a bug where saving as a PDF would make all the other buttons open in a new tab.

**See Screenshot Below:**
> Screen Shot 2019-09-02 at 1 56 03 pm

### DIFF
--- a/cassdegrees/static/js/student_edit.js
+++ b/cassdegrees/static/js/student_edit.js
@@ -200,12 +200,43 @@ interact('.dropzone').dropzone({
 
 
 // When selecting a sublan from a list of subplans, sets the chosen subplan as the saved value
-function subplanSelect(element){
+function subplanSelect(element, year){
     var inputField = element.parentNode;
     while (inputField.tagName !== 'INPUT') {
         inputField = inputField.previousSibling;
     }
     inputField.value = element.attributes['name'].value;
+
+    // Set the box values
+    var code = inputField.value.split(' ', 1);
+    var request = new XMLHttpRequest();
+    request.addEventListener("load", function() {
+        var rule = JSON.parse(request.response)[0]["rules"];
+        console.log(rule);
+        var card = inputField.parentNode.nextSibling;
+        while (card.tagName !== 'DIV')
+            card = card.nextSibling;
+        for (var i=0; i<rule.length; i++){
+            if (rule[i]["codes"].length*6 === parseInt(rule[i]["unit_count"]) && rule[i]["list_type"] === "exact"){
+                for (var j=0; j<rule[i]["codes"].length; j++){
+                    var course = rule[i]["codes"][j];
+                    card.childNodes[0].setAttribute("data-course-code", course);
+                    card.getElementsByClassName("course-code")[0].innerText = course;
+                    card = card.nextSibling;
+                    while (card.tagName !== 'DIV')
+                        card = card.nextSibling;
+                }
+            }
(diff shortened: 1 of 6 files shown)


aidengeorge97 (approved):
Works well, and bringing everything on the page into the form is a good solution. I encountered no issues testing it.

cassdegrees/static/js/student_edit.js
```diff
@@ -200,12 +202,43 @@ interact('.dropzone').dropzone({
...
         inputField = inputField.previousSibling;
     }
     inputField.value = element.attributes['name'].value;
+
+    // Set the box values
+    var code = inputField.value.split(' ', 1);
+    var request = new XMLHttpRequest();
+    request.addEventListener("load", function() {
+        var rule = JSON.parse(request.response)[0]["rules"];
+        console.log(rule);
```
j-selby: Unneeded logging

cassdegrees/static/js/student_edit.js
```diff
@@ -200,12 +202,43 @@ interact('.dropzone').dropzone({
...
+        var card = inputField.parentNode.nextSibling;
+        while (card.tagName !== 'DIV')
+            card = card.nextSibling;
+        for (var i=0; i<rule.length; i++){
+            if (rule[i]["codes"].length*6 === parseInt(rule[i]["unit_count"]) && rule[i]["list_type"] === "exact"){
+                for (var j=0; j<rule[i]["codes"].length; j++){
```
j-selby, line 225: This is fairly messy. Can this be cleaned up (spacing + comments)?

cassdegrees/ui/templatetags/student_course_boxes.py
```diff
@@ -49,5 +49,5 @@ def index(List, i):
 
 
 @register.filter
-def listIndex(List, i):
-    return List[i]
+def get(List, i):
+    return List[str(i)]
```
j-selby, line 53: `str` or `int` here? If it is a string key maybe this should be specified as something like `get_by_string`?

cassdegrees/static/js/student_edit.js
```diff
@@ -200,12 +202,43 @@ interact('.dropzone').dropzone({
...
+            }
+            else{}
+        }
+        while(card.nextSibling) {
+            if (card.tagName === 'DIV')
+                clearCourse(card.childNodes[0]);
+            card = card.nextSibling;
+        }
+    });
+    request.open("GET", "/api/search/?select=code,year,rules&from=subplan&year="+year+"&code="+code);
```
j-selby: This could in theory allow for XSS - perhaps sanitise year?

j-selby (changes requested):
Looks good, but needs a few tweaks.

cassdegrees/ui/templatetags/student_course_boxes.py
JackSekoranja: Yeah Django was being very strange here. Even when the `List` dictionary had a integer key, it would be converted to a string anyway and would break because it couldn't find the integer key. The reason this worked before was that `List` was an array rather than a dictionary, so ints were valid. I named it `get` to reference the fact that python's `get` function is only used in dictionaries. If you still think I need to be more specific then I will, but I think this is fine

PR #303, opened 2019-09-03
Fixed breadcrumbs on create list page

aidengeorge97:
Simple bug fix to redirect the breadcrumbs on the create list page to staff url rather than admin.


JackSekoranja (approved):
Tested, happy to merge

PR #302, opened 2019-09-03
Code cleanup/bugfix: remove misc 

j-selby:
This PR removes mistakenly inserted migration files, and strengthens the git ignore file to prevent erroneous files (database, caches) from being added.

PR #305, opened 2019-09-03
Code cleanup: Refactor components into modules

j-selby:
Depends on #302.

This PR refactors scripts, templates, views and so forth into submodules so that they can be easily identified. This solves the issue where identifying the split between staff and student systems might not be immediately obvious.


JackSekoranja (approved):
Tested everything I could think of and it seems to work fine. I think I'd prefer for this PR to be pushed after all current ones get through to avoid merge issues

j-selby:
All other PRs have been merged that are going to be merged (except #309, which is similarly big), so I'll rebase this one shortly.

Issue #311, opened 2019-09-04
Update Bulk Uploader to work with the updated course model


Daniel-Jang:
Currently, the bulk uploader works with the old models. It will break with the revamped course models, so fix this to make it work with the new course models.

PR #309, opened 2019-09-04
Streamline course model

Daniel-Jang:
Original Issue: #310 

This is the new model for courses:
image

Summary:
- Replaced year with offered years (odd, even etc)
- Added support for sessions (summer, winter etc)
- Added support for 'other' offerings (generic catch-all)
- Added 'currentlyActive' flag to let users know if this course is active
- Updated the entire code base to remove usage of 'year'.

No new features in this PR, but mostly going around and removing all usages of 'year'.
The 'currentlyActive' flag doesn't do anything at the moment (e.g. prevent inactive courses from being added into programs), but that should be part of a future issue/PR.

If any of you find any bugs, please let me know.

PR #315, opened 2019-09-05
Update bulk uploader to work with the new Course Model

Daniel-Jang:
Issue: #311 

image

Now that the course model has completely changed, I have modified the bulk uploader to support uploading of '% separated values' and updated the code that processes the CASS provided Excel file.

Here is a summary of the changes:
- Updated bulk uploader to require input for sessions offerings and other flags such as 'active'.
- Updated code to process the excel file provided by CASS without errors.
- Improved S1, S2 offering detections
- Added support for sessions (winter, summer sessions)
- Added support for courses with strange offerings (e.g. S1 in 2019, S2 in 2020), where the 'otherOffering' value is set to be true to mark it as 'strange' and 'offeredYears' is set as "OTHER". Previously, these would be ignored.
- Does not add courses which are marked for 'disestablishment'


cassdegrees/api/models.py
```diff
@@ -11,16 +11,34 @@ class SampleModel(models.Model):
...
+    otherOffering = models.BooleanField()
+
+    currentlyActive = models.BooleanField()
+
     rules = psql.JSONField(default=list)
     lastUpdated = models.DateField(default=timezone.now)
 
     class Meta:
-        unique_together = (("code", "year"),)
+        unique_together = (("code", "name"),)
```
j-selby, line 41: Not sure about this one - is there any value for name also being unique?

cassdegrees/ui/views/staff/bulk_data_upload.py
```diff
@@ -182,22 +220,38 @@ def bulk_data_upload(request):
...
         # so that the right data is stored in the right column
         # This would also allow columns to be in any order, and courses/subplans would still be added.
-        map = {}
+        map = dict()
         for row in uploaded_file:
             if first_row_checked:
                 if content_type == 'Courses':
                     # If number of columns from file doesn't match the model, return error to user.
-                    if len(row) != 6:
+                    if len(row) != 11:
```
j-selby, line 228: Is it possible to not hard code this one? Not exactly your problem, but yeah.

j-selby (approved):
Looks good!

cassdegrees/api/models.py
Daniel-Jang: According to Jack, CASS recycles course codes so there will be different courses with same course code. 
Oh and also, this belongs in the other pull request:
https://github.com/cass-degrees/CASS-Degrees-Code/pull/309

cassdegrees/ui/views/staff/bulk_data_upload.py
Daniel-Jang: Unfortunately no, this is one of the checks to make sure the file format is in a desirable format.

Issue #261, opened 2019-08-26
Subplan Selection Doesn't Distinguish Majors, Minors or Specs


JackSekoranja:
When creating Programs, selecting the "Subplan" option will give a list that doesn't differentiate Majors, Minors or Specialisations. This shows a lot of unnecessary information to the user, as they can only select a plan of a single type. Furthermore, Minors and Specialisations have the same unit count, so the user wouldn't be able to differentiate between the two. To fix, it would be great for the user to be able to select a specific subplan type, and for only those to appear in the list.

![Screen Shot 2019-08-26 at 6 03 26 pm](https://user-images.githubusercontent.com/36946090/63675502-4d269080-c82c-11e9-9cd7-53b15e780f2c.jpg)

JackSekoranja:
Closed with PR #323

PR #323, opened 2019-09-06
Modify the subplan rule to ask for the subplan type

JackSekoranja:
This PR closes issue #261.

Users will now be prompted for a subplan type and the list below it will dynamically change based on this selection. Subplan rules also now save a `subplan_type` which is currently unused in other areas of the application, but was left in as it could still be useful.

Screen Shot 2019-09-06 at 5 00 59 pm


j-selby (approved):
Looks good!

Issue #239, opened 2019-08-23
Courses with dependencies prevents deletion of other marked courses


Daniel-Jang:
During deletion, if even 1 course has a dependency issue, all the other courses that were marked for deletion won't be deleted.

![image](https://user-images.githubusercontent.com/37033052/63572549-75ad5100-c5c6-11e9-982f-eacb7d019f88.png)

All the courses that don't have dependency issues should be deleted.

JackSekoranja:
Closed with PR #324

PR #324, opened 2019-09-06
Bug fix dependency deletion overwrite

JackSekoranja:
This PR closes issue #239.

Now if the user tries to delete a set of courses where some are invalid, they will be notified and asked if they want to delete the remaining ones.

Screen Shot 2019-09-06 at 5 47 02 pm

### DIFF
--- a/cassdegrees/static/js/staff/rules.js
+++ b/cassdegrees/static/js/staff/rules.js
@@ -77,6 +77,12 @@ const LIST_TYPES = {
     'max': 'No more than'
 };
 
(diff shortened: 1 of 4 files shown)


j-selby:
This PR also seems to include changes from #323

JackSekoranja:
> This PR also seems to include changes from #323

Ah, I must have branched off the wrong branch. I'll just merge this one after #323

j-selby (approved):
Looks good otherwise! (would need to wait for #323 however)

PR #329, opened 2019-09-09
Course creation on forms without changing page

### LINKED ISSUE #284: Allow users to either add custom courses and/or have an easy way to add new courses
Currently, the user can't add courses that isn't uploaded. If the course the user is looking for is not available, the user has to save and exit, add the course in question and re-edit the program/subplan again.

Rather, we should have a way for the users to add custom course right on the spot and/or a shortcut to add a course, like we do with subplans when creating programs.

Link to UAT: https://docs.google.com/document/d/1KfdzBN3GWYagc18__njZcWjA_km5c015UwtlQM5y_ww/edit#

JackSekoranja:
This PR closes #284 

Now users are able to create new courses (without specifying requisites) without leaving the current form. It also works off the course model we specified in Django, and should therefore change with the original form.

**Blank Form With More Button Active:**
Screen Shot 2019-09-09 at 10 46 04 pm

**Form With Errors:**
Screen Shot 2019-09-09 at 10 46 42 pm

**Form After Success:**
Screen Shot 2019-09-09 at 10 48 57 pm


JackSekoranja:
After talking with @Daniel-Jang I'm going to see if I can get this working without JQuery/AJAX

JackSekoranja:
Relevant changes have been made, please review as normal

j-selby:
Does this duplicate regular course creation? Is it possible to even frame that page?

JackSekoranja:
> Does this duplicate regular course creation? Is it possible to even frame that page?

Stylistically it does, although it hides most of the scheduling information as we’ve made that a lower priority. The requisite (vue) components are completely removed so this isn’t supposed to be a replacement for the course creation page, it’s just trying to make it easier to quickly create courses while working on a plan

Daniel-Jang:
While I have no issues with the 'show more' option, I think you should include the "offered in semester 1 and 2" checkboxes in the window in the 'show less' view.

JackSekoranja:
Yeah, I decided against that as the clients said they wanted to focus more on the plan structure rather than scheduling. As semesters is a scheduling element it has been hidden

Daniel-Jang:
Ah okay. We can leave this PR as-is and we can also get feedback from our clients tomorrow about this.

Daniel-Jang (approved):

PR #330, opened 2019-09-10
Subject Area Global Requirements

### LINKED ISSUE #259: Add Subject Area Global Requirements
We are currently unable to add global requirements referring to subject-area courses without putting it in a custom text field as part of the program (Which may confuse our system or other students). I think we need to have a dedicated rule for this in global requirements.

Consider the following from a *Bachelor of Music*
 (BMUSI - [Found Here](https://programsandcourses.anu.edu.au/program/BMUSI)):
> "A minimum of 84 units must come from completion of courses from the subject area MUSI Music and courses from any majors listed in the Bachelor of Music."

Currently the best known approach is to add it in the fields below. See the following screenshot for reference:
Screen Shot 2019-08-26 at 5 44 57 pm

aidengeorge97:
Closes #259.

This PR has merged the existing global requirements into one generalised rule. This serves the purpose of reducing confusion over the titles of global requirements which users seemed to have in UAT. It also enables more complicated global requirements to be created with subject area constraints and streamlines the creation process.

image

image

image

### DIFF
--- a/cassdegrees/static/js/staff/globalrequirements.js
+++ b/cassdegrees/static/js/staff/globalrequirements.js
@@ -89,7 +89,12 @@ Vue.component('global_requirement_general', {
         check_options: function() {
             this.invalid_units = this.details.unit_count <= 0;
             this.invalid_units_step = this.details.unit_count % 6 !== 0;
-            this.is_blank = this.details.unit_count === "";
+            this.units_is_blank = this.details.unit_count === "";
+
+            this.is_blank = !this.details.courses1000Level && !this.details.courses2000Level && !this.details.courses3000Level
+                && !this.details.courses4000Level && !this.details.courses5000Level && !this.details.courses6000Level
+                && !this.details.courses7000Level && !this.details.courses8000Level && !this.details.courses9000Level
+                && this.details.subject_area === "any";
         }
     },
     template: '#generalGlobalRequirementTemplate'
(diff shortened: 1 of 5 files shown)


JackSekoranja (changes requested):
Found a few bugs/issues I'd like fixed. There is currently an issue where the subject area box won't always update. This allows people to store bad data in the database as follows:
1. Add new Global Requirement
2. Set subject area to anything other than "any"
3. Set units to 6
4. Set subject area to "any"
5. Save

This also doesn't fully address #259, as it doesn't address the provided example, as the "any courses from any majors listed in the Bachelor of Music" can't be represented:
>"A minimum of 84 units must come from completion of courses from the subject area MUSI Music and courses from any majors listed in the Bachelor of Music."

To fix this, maybe a general text-based rule should also be added

aidengeorge97:
I identified a bug with the existing system where global requirements were not being validated before form submission. This may have contributed to what you observed Jack, as I believe the error messages work fine but the form could still be submitted regardless. I added in validation for global requirements and added the custom rule, so this should meet your requests now.

Daniel-Jang (approved):
I've played around with this and I think it works well now, especially after following Jack's suggestions.

JackSekoranja (approved):
Okay, tested and it works. I do think the rule is a little complicated now so it would be great if we could simplify it somehow, but I'll leave that for you to decide

JackSekoranja:
Closing on Aiden's behalf due to upcoming UAT

PR #331, opened 2019-09-10
Improve readability of form elements on screen

j-selby:
This PR modifies form elements throughout the site to be be more readable.

Overarching changes made:
- Grey backgrounds removed where possible - these can potentially imply that a element is disabled, and may make prompt text illegible in low-contrast situations. This was tested with a low-contrast filter to ensure that this is now usable.
- Sizes of most elements (excluding checkbox, see below) have been increased to prevent overcrowding of input areas.
- Borders of most input elements have been darkened to differentiate them from the rest of the page. These were tested with various colour blindness filters.
- Top navigation link (CASS Program Planner) now has a underline to indicate interactivity.
- Buttons have improved contrast and hover behaviour.
- Light coloured text in the page has been darkened.

Before:
before

After:
after


Daniel-Jang (approved):
I had a look and feel of it, and I don't have any issues with it. Looks good to me.

JackSekoranja (approved):
Looks much better and is definitely easier to use. I think the checkboxes should be a bit bigger as Andrew said in the UAT that he thought they were small, but I still think this is a massive improvement

JackSekoranja:
Merging this on James' behalf due to the upcoming UAT

PR #336, opened 2019-09-11
Implement custom checkboxes globally

j-selby:
This took *way* too long.

This PR implements themed checkboxes to improve clarity in the overall forms. The initial problem faced by this was that checkboxes *cannot* be styled by CSS in most browsers - they always use the platform checkbox, and that is all you get. The workaround, in this case, is to completely reimplement the rendering of checkboxes (using a library - not all in this PR!) and do that rendering ourselves.

The second issue is that this custom rendering approach requires additional elements around the checkbox to facilitate custom styling (again, noting that CSS doesn't allow for this). Django, however, doesn't allow for this by default when compositing forms - we would need custom logic for this, which I deemed too overkill for this.

In this PR therefore, I implement a custom JavaScript event handler on page loads which attempts to swap out any checkboxes it can find with styled ones, falling back gracefully if this is not possible. This doesn't work on Vue elements - these have to be manually styled.

Screenshot:

checkbox


JackSekoranja (approved):
It’s a shame there isn’t an easier way to do this but it does work. I did notice that the /staff/list page doesn’t render the special checkboxes but that should be okay since they’re not really supposed to be deleting courses anyway. Happy for you to merge

aidengeorge97 (changes requested):
Looks good but it I'm getting some additional artefacts appearing on/around the boxes? This might just be my set up but I'd like to get your thoughts before I approve the PR.

Google Chrome:
image

Microsoft Edge:
image

spencerj411 (commented):
I'm also experiencing what Aiden got
**On Google Chrome:**
image

j-selby:
Issues with Chrome/Edge should now be fixed - they were both behaving weirdly when dealing with overflow.

aidengeorge97 (approved):
Those changes seem to have fixed it. Happy to approve now.

PR #337, opened 2019-09-12
Added Visual Cues to Rule Creation

### LINKED ISSUE #322: Visual Cue when a New Rule is Added
UAT revealed that users sometimes do not see where a new rule has been created. For example, Justine creates a new rule, which appears at the bottom of the screen, Justine does not notice the new rule and attempts to edit a previous rule mistaking it for the new one. A visual cue which clearly indicates to the user when/where a new rule has been created would solve this.

Suggestions:
* Scroll the page down so the new rule is centered
* Show a flash of colour around the box that fades
* Display a solid bar to the right of the new rule which fades or stays until the next new rule.

aidengeorge97:
Closes #322. 

The purpose of this PR is to make it clear when and where a new rule has been added into a program/subplan/course template. Now when a rule is added, the page will automatically scroll to it and perform a fade-in effect.

I'm undecided on the scroll effect currently so I'd like some feedback. You can delete line 135 in globalrequirements.js to see how it looks without the scroll when adding global requirements. It may be a little intrusive, but it ensures the new rule will be visible on screen.


spencerj411 (approved):
Had a mess around, and looks good to me! I personally like the transition, though it is pretty choppy (could just be on my laptop); sometimes, even more so when the browser window is smaller. Apart from that, I'm happy for you to merge.

j-selby:
I'm not sure a fade is the best option here, especially if the scrolling animation itself might disorient the user - is it possible to instead look at something like this to aid with general navigation as well?

left-bar

For reference, this was implemented with:
```
border-left: 2px solid green;
margin: 0 0 0 -6px;
padding: 0 0 0 4px;
```
on the parent element of a `rule-container`.

aidengeorge97:
Hmm that's a fair thought. I've pushed a commit which replaces the fade with your side bar just for global requirements. Could I get some thoughts before I decide whether to implement it completely?

JackSekoranja (commented):
I found that you can break this slightly by doing the following:
1. Add an OR Rule
2. Add a rule inside the OR Rule
3. Add a rule below the OR Rule
4. Add another rule in the OR rule

This results in the green bar staying on the rule created in step 3. I don't think it's a huge issue but it would be nice to fix

aidengeorge97:
> I found that you can break this slightly by doing the following:
> 
> 1. Add an OR Rule
> 2. Add a rule inside the OR Rule
> 3. Add a rule below the OR Rule
> 4. Add another rule in the OR rule
> 
> This results in the green bar staying on the rule created in step 3. I don't think it's a huge issue but it would be nice to fix

Yea, because our current way of updating the display is by deleting and then recreating all the vue components it's hard to detect which rule has just been added. I decided to assume that the bottom most rule will generally be the current one so that's where the cue goes which. I can try keep working on it if you want it perfect though?

JackSekoranja:
> > I found that you can break this slightly by doing the following:
> > 
> > 1. Add an OR Rule
> > 2. Add a rule inside the OR Rule
> > 3. Add a rule below the OR Rule
> > 4. Add another rule in the OR rule
> > 
> > This results in the green bar staying on the rule created in step 3. I don't think it's a huge issue but it would be nice to fix
> 
> Yea, because our current way of updating the display is by deleting and then recreating all the vue components it's hard to detect which rule has just been added. I decided to assume that the bottom most rule will generally be the current one so that's where the cue goes which. I can try keep working on it if you want it perfect though?

Na you don’t need to worry about that. I was just pointing it out as if you add a rule at the bottom and then one in an OR rule the newer one will be marked despite it not being at the bottom. I figured if you had that functionality then you were going off the newly added rules

Daniel-Jang (approved):
I think it looks simple and nice.

aidengeorge97:
After much revision I believe this feature is ready for final review and merge. Now the most recently created rule will be highlighted with a green bar, and the page will scroll down to it.

To apply the effect to the most recent rule I had to modify the redraw functions used to refresh Vue components, so this PR also falls under Code Quality. Now Vue will no longer delete and redraw all rules just to redraw one, only the updated rule will be modified.

JackSekoranja (approved):
Works well, happy for you to merge

Issue #296, opened 2019-08-31
Allow the user to copy certain aspects of certain rules (e.g. copy or rule group)


Daniel-Jang:
UAT Link: https://docs.google.com/document/d/1KfdzBN3GWYagc18__njZcWjA_km5c015UwtlQM5y_ww/edit#

Allow the user to copy and paste rules so that they don't have to punch in the same rules over and over again.

PR #340, opened 2019-09-12
Feature Duplicate Rules

Daniel-Jang:
Original Issue: #296 

Allows the user to duplicate any rule they wish to duplicate. Also, the user can duplicate an or rule group.

Before:
image

After:
image

Or rule group
Before:
image

After:
image


cassdegrees/static/js/staff/rules.js
```diff
@@ -831,10 +831,20 @@ Vue.component('rule_either_or', {
             this.details.either_or[group].splice(index, 1);
             this.do_redraw();
         },
+        duplicate_rule: function(index, group) {
+            // JSON.parse(JSON.stringify(...)) is done to actually duplicate the contents of the rule, rather than just copying the memory references.
+            this.details.either_or[group].splice(index, 0, JSON.parse(JSON.stringify(this.details.either_or[group][index])));
```
j-selby, line 852: JSON encoding/decoding is ugly here, but this is JavaScript, so its all ugly anyway :P - does the job anyway.

j-selby (changes requested):
This doesn't appear to work on the courses page - creating a rule and pressing duplicate does nothing, incl. in the console. (Firefox 71.0a1, Windows 10).

Daniel-Jang:
That is interesting, I will fix that, thanks James.

cassdegrees/static/js/staff/rules.js
Daniel-Jang: HAHA yeah, I did not like this solution but there was no better way...

JackSekoranja (approved):
Works well, happy for you to merge.

j-selby (approved):

PR #343, opened 2019-09-13
Empty Dropdowns When No Course/Subplan/Program in Database

### LINKED ISSUE #338: Empty Dropdown Menus When There are No Courses/Subplans/Programs
An empty dropdown menu shows up (see below) when there are no programs/subplans/courses in the database.
_Different cases:_

- **adding a subplan rule to a program**

image

- **adding a course/program requisite to a course** 

image

- **adding a course list rule to a subplan**

spencerj411:
closes #338.

When a user is trying to add a rule/requisite/incompatibility that depends on say, courses, if there are no courses in the database, a message will pop up notifying the user that they need to created courses first before they can create the rule/requisite/incompatibility they're planning to create. This was implemented with all the rules/requisites/incompatibilities in mind (using vue.js).
Eg.
**No Programs**
image
**At least one Program**
image

### DIFF
--- a/cassdegrees/static/js/staff/rules.js
+++ b/cassdegrees/static/js/staff/rules.js
@@ -85,12 +85,12 @@ const SUBPLAN_TYPES = {
 
 const INFO_MSGS = {
     'course': '<p>This Requisite requires Courses in the system. Please create Courses ' +
-        '<a href="/staff/create/course/">here</a> or bulk upload Courses ' +
-        '<a href="/staff/bulk_upload/">here</a> first before creating this Requisite.</p>',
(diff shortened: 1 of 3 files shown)


KavsHub (approved):
Looks good. I tested subplans and courses too.
I also noticed the bug #344 that you raised too haha.

j-selby (changes requested):
Looks good, but this appears to discard a users progress in creating a course/... when following these links? Should this be a warning or open in a different tab (`target='_blank'`) or something?

j-selby (approved):
Issue fixed. Nice work!

PR #351, opened 2019-09-15
Implement confirmation box for saving non-public programs/subplans

### LINKED ISSUE #321: Provide a Notification when Saving an Unpublished Template
UAT revealed that users may pass over the publish button and become confused when the item they just created is not visible elsewhere. It would be helpful to show a little notification to the user when they are saving a plan without publishing it, informing them that this plan will not be visible elsewhere. For extra points, this notification could have a "Do not show me this again" check box.

j-selby:
Closes #321.

This PR implements a simple confirmation box on the submission of programs/subplans to confirm the user's decision to save the respective plan as non-public. A browser confirmation was chosen as it is both blocking (the window is non-interactive otherwise) as well as mutable with built-in ways to manage that.

Screenshot (Firefox 71.0a1):
public-notification

### DIFF
--- a/cassdegrees/templates/staff/creation/createprogram.html
+++ b/cassdegrees/templates/staff/creation/createprogram.html
@@ -111,6 +111,13 @@
 
         function submit_form(form_action, redirect) {
             if (handleProgram() && handleRules()) {
+                // Ensure that the course has been agreed to to being non public on submission.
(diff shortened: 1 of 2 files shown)


JackSekoranja (approved):
Tested and it works well, happy for you to merge

Daniel-Jang (approved):
Seems good to me.

PR #352, opened 2019-09-15
Verify that plan links exist when setting up a popup

### LINKED ISSUE #304: Popup code throws a null error in console if no popup defined
The popup code appears to throw errors if no popup is defined by the runtime. This should be suppressed if this is intentional, or should be fixed if its an actual bug.

This was tested on Firefox 70.0a1, Windows 10.

```
TypeError: plan_link is null @ edit.js:55:5
    setupPopup  http://127.0.0.1:8000/static/js/edit.js:55
    <anonymous> http://127.0.0.1:8000/edit/?plan=blah:12273
```

j-selby:
Closes #304.

This adds a safety check to ensure that popups on the student facing system (e.g. for generating a sharing link) are set up correctly, and their handlers aren't injected if their respective elements don't exist.

This already appears to be masked by #298, but for sake of safety a type check has been added to ensure that this isn't an issue even if this does end up getting called during execution.

### DIFF
--- a/cassdegrees/static/js/student/edit.js
+++ b/cassdegrees/static/js/student/edit.js
@@ -55,6 +55,7 @@ function setupPopup() {
         return;
     }
 
+    // Plan link implies these:
     var copy = document.getElementById("copy_to_clipboard");
     var close = document.getElementById("close_modal");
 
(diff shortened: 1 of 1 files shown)


JackSekoranja (approved):
Wasn't sure how to test this one, but the code for it looks good and the popups seem to work. Happy for you to merge

Daniel-Jang (approved):
Sure, it looks good.

Issue #342, opened 2019-09-13
Cancel Button in Create Course/Subplan/Program Links to Staff Home Page


spencerj411:
When you Edit a Course/Subplan/Program, and click the Cancel button at the bottom, it links you back to the _Program, Subplan, and Course List_ page. However, when you **Create or Duplicate** a Course/Subplan/Program, clicking the Cancel button will link you back to the staff homepage. If this is intentional, please delete this issue.

KavsHub:
I feel like for consistency we should link back to where they came from previously.

However, I think the reason why they go back to the homepage is because the create program/subplan/course main page's cancel button's redirect to home if you press cancel.
I'll see if I can edit the cancel buttons to return to the corresponding list page if the user pressed create on the list page not the main page.

KavsHub:
Fixed as of PR #353

PR #353, opened 2019-09-15
Fixed cancel button redirects

KavsHub:
Issue #342 

Cancel Button will now redirect to the relevant lists page if the user came from the lists page.
The button will still redirect to the main page if the user pressed the create buttons on the main page.

This was simply done by checking if the URL the user came from was the lists page. This means if the user types the URL or comes from somewhere else it will still default to the homepage if not the lists page.

### DIFF
--- a/cassdegrees/templates/staff/creation/createcourse.html
+++ b/cassdegrees/templates/staff/creation/createcourse.html
@@ -63,7 +63,7 @@
 
     <p class="text-right">
         <input class="btn-uni-grad btn-large" type="button" value="Cancel"
-               onclick="{% if edit %}returnToList('Course'){% else %}goBack(){% endif %}">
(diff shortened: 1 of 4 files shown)


j-selby:
I'm not sure if using the referrer is the best option here - browsers may elect to not send this (particularly if private browsing or ad blockers are enabled). Would `history.back()` or similar work instead?

j-selby (changes requested):
Solution otherwise looks fine - see above comment.

KavsHub:
@j-selby We could use history.back() but if the user didn't come from list then they would return to where they came from i.e. if I was at www.google.com then I typed url/staff/create history.back() would take me to google.com instead of the homepage (which it defaults to).

This is why I opted for checking if lists was in the referrer url and if lists wasn't in the url then it would return to the homepage by default (solving the google.com example from above).
This does however, mean that if the referrer url had lists in its name (www.ilovelists.com or something) then it would also return to the lists page (which is fine as it is still our website)

PR #354, opened 2019-09-15
Feature Delete Button in Edit

### LINKED ISSUE #289: Add Delete Feature in Edit Course/Subplan/Program

spencerj411:
closes #289.
This pr add a new delete button in Edit Course/Subplan/Program (none for List as it currently does not have a delete feature).
Thoughts on the size of the button (I used the `btn-small` class) and the wastebasket/rubbish bin char?
**edit.html**
image

**list.html**
image


JackSekoranja (approved):
Works on everything that I have tested. I'm not sure about the use of the emoji but I'll let other members decide. If you did want an icon we could probably look in to vector graphics as we have done with the main menu

PR #362, opened 2019-09-17
Implement several security checks for API endpoints

j-selby:
This PR implements several security checks to ensure that unauthenticated users (defined as those who are not logged in as staff) cannot make any modifications to the database nor can they access non-published plans and subplans.

This is implemented through three mechanisms:
- The search interface explicitly checks for authenticated users and adds a forced parameter of `publish=true` if they are not. This overrides any other filters on `publish`.
- A middleman has been created for API views which checks for users identities and hotswaps out the models exposed to users via a filter to public plans only if users are not logged in.
- REST Framework's permissions system has been configured to only allow write changes for authenticated users.

This should require 0 code changes elsewhere in the codebase, and has been tested in many scenarios, but more testing might be needed.

### DIFF
--- a/cassdegrees/api/views.py
+++ b/cassdegrees/api/views.py
@@ -139,7 +139,8 @@ def search(request):
     )
 
     # Ensure only authenticated users can access secret plans
-    if not request.user.is_authenticated and hasattr(model, "publish"):
+    # Request.user will not exist if this is an internal request - thats perfectly fine, however!
+    if request.user and not request.user.is_authenticated and hasattr(model, "publish"):
         include["publish"] = True
 
     query = Q(**include)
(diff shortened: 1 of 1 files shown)


JackSekoranja (changes requested):
Keep getting this issue where trying to view a program/subplan or generate a PDF. It also prevents the user from deleting Subplans and Courses. The issue always seems to be that the new security measures don't work with the old ones, as it always seems to be an error with `request.user`.

All of the aforementioned errors are similar to this one:
Screen Shot 2019-09-17 at 7 07 34 pm

j-selby:
@JackSekoranja that should be fixed now

JackSekoranja (changes requested):
Still having the same issues when creating PDFs, viewing programs and subplans, and deleting courses, etc.

JackSekoranja (approved):
Seems to work fine now :), happy to merge

aidengeorge97 (approved):
Looks good and works as expected. Happy for this to be merged.

PR #364, opened 2019-09-17
Uat list integration

joshcbee:
Initial implementation and integration of new list UI. Currently lists are mutable in rule creation, and previously created lists can be used to batch add courses. This implementation does not implement list naming (to be added in a follow-up issue) and backward creation of lists from the UI tool (to be implemented in a future release if needed).

image

image


JackSekoranja (approved):
Looks good and works really well so I'm happy for you to merge. 

I think there are a few stylistic changes that could be made if possible. When adding more than 8ish courses they run off the screen so users are unable to see what they have added, so it might be good if the courses appeared above the text box. I also think the spaces between added courses are quite large and take up a lot of the screen as a result, so I think it would be better to remove some of the space between them. I also think having a more clear indicator that the courses come from the list would be a good idea as currently it looks like they are just floating below the text box. Again, would be great to get to these at some point if we could but for now it seems fine.

aidengeorge97 (changes requested):
This looks good for the programs page, but on the subplan page I hit an error which seems to have disabled the ability to add course rules. 

image

I'd also like to see this feature added to the course requisite and course incompatibility rules in courses for consistency. I also agree with Jack's points on the style changes.

joshcbee:
Subplan issue has been resolved - we have duplicate templates for subplanrules.html that we may want to refactor as part of a future issue?

Also added minor styling changes:
- reduced space between list items
- added slight indent to left of items
- remove top margin of list

Re styling: I think we just need to be careful of the trade off between having to scroll a bit less and creating too much visual clutter. Happy to continue to review though.

Will be rolling out UI change to requisites and incompatibilities as part of rollout to subplans issue.

image

aidengeorge97 (approved):
With those changes I'm happy to approve now. I think a potential feature for a future issue would be the ability to collapse/expand the long lists of courses to ease navigation.

PR #366, opened 2019-09-17
Unique Course Rule Identifiers

### LINKED ISSUE #266: Program View - All Courses with the Same Code are Shown
If there exist multiple versions of the same course code, but with different names, then the view page will display all courses with that code, not just the one that was actually selected.

Example. The course rule is defined like this:

image

But displays this on the view page.

image

aidengeorge97:
Closes #266.

This PR actually ended up addressing a large issue to solve the minor view issue as described in the original issue.

The existing model only saves the course codes for a list of course requirements (or requisites or incompatibilities), which means that if there are two courses with the same code and different names (which our interface allows), then the courses will be conflated. 

If you create two courses with the same course code and different names and then attempt to select both rules in a course rule, you will find you can only select one. And when you view the program it will display both rules when only one was selected.

This PR 'simply' updates it to store the code and name of courses for such rules, and refactors the program to account for this. I've updated the pointers in PDF generation, view pages, other rules, and the student website, but it's possible I may have missed one.

### DIFF
--- a/cassdegrees/static/js/staff/rules.js
+++ b/cassdegrees/static/js/staff/rules.js
@@ -506,7 +506,7 @@ Vue.component('rule_course', {
             if (!(rule.details.codes.length === 0)) {
                 for (let i = 0; i < rule.details.codes.length; i++){
                     for (let x = 0; x < rule.courses.length; x++){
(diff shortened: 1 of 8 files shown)


JackSekoranja (approved):
Seems to work well, happy for you to merge

PR #369, opened 2019-09-18
Uat delete lists

joshcbee:
Implementation of list deletion.

Expected functionality: allow a user to delete lists contained in the database. On deletion, these lists should no longer be available for selection.

Out of scope: deletion results in the deletion of the list from plans which may contain it. Lists are mutable collections of courses use to bulk add requirements to plans. Plans themselves do not contain a reference to the lists, and have their own local collections of courses.

image

image


JackSekoranja (approved):
Tested and it works. Happy for you to merge

Issue #370, opened 2019-09-18
Error When Bulk Uploading Invalid File Types


aidengeorge97:
The bulk upload feature should perform a check to ensure that the chosen file type is valid and return a corresponding error message.

When uploading an mp4 file:

![image](https://user-images.githubusercontent.com/37424867/65119119-3809df80-da2f-11e9-8cea-ca68ace3587c.png)

KavsHub:
Closed by PR #375

PR #375, opened 2019-09-18
Checking filetype of uploaded file for bulk uploading

KavsHub:
Addressed the issue [Error When Bulk Uploading Invalid File Types] #370 

The issue was caused by TextIOWrapper function not accepting some encodings like images and videos.

Fixed this by checking the file extension before we parse the file into the wrapper. 
Since there is sanity checking later in the code, there isn't a need to verify the contents since its already done.

This was a simple fix to the issue. I am aware that if someone named an image file .xls or csv it could potentially still crash the wrapper - however, I don't believe CASS is this malicious so its not worth the effort as of now.

### DIFF
--- a/cassdegrees/ui/views/staff/bulk_data_upload.py
+++ b/cassdegrees/ui/views/staff/bulk_data_upload.py
@@ -24,6 +24,7 @@ from django.shortcuts import render
 # the % sign in a new cell in that row).
 
 # Support is available for CASS' custom teaching plan excel file.
+import os
 
 
 @login_required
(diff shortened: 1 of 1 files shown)


cassdegrees/ui/views/staff/bulk_data_upload.py
```diff
@@ -41,6 +42,18 @@ def bulk_data_upload(request):
         # Open file in text mode:
         # https://stackoverflow.com/questions/16243023/how-to-resolve-iterator-should-return-strings-not-bytes
         raw_uploaded_file = request.FILES['uploaded_file']
+        file_type = os.path.splitext(str(raw_uploaded_file))[1]
+
+        # Put the supported file extensions in this list
+        supported_file_types = {"xlsx", "xls", "csv"}
```
JackSekoranja:
```suggestion
        supported_file_types = [".xlsx", ".xls", ".csv"]
```

JackSekoranja (changes requested):
Didn't work for me, I needed to make the following change for it to work.

cassdegrees/ui/views/staff/bulk_data_upload.py
KavsHub: My bad, you're right.

PR #378, opened 2019-09-19
Fix "Create Subplan" Button From Program Edit 

### LINKED ISSUE #258: 'Create new subplan' button doesn't work if the program has errors
When creating a program, the user is unable to click the "add new subplan" button if there are errors anywhere in the form. Since they may be in the process of adding the subplan list when they want to click it, this is something we should fix

aidengeorge97:
Closes #258.

Simple little fix to allow users to create a subplan from the create/edit program page, and also consolidates an existing function.


JackSekoranja (approved):
Looks good, happy for you to merge.

Issue #365, opened 2019-09-17
List feature currently broken


JackSekoranja:
There currently seems to be a lot of issues surrounding the recently added "Lists" feature. Clicking "View" on the lists section doesn't work, so we will need to define a "view" option for lists. Also if a new list has the same name and year as an existing one it will crash.

joshcbee:
Can you check if this is still happening with in the PR branch, there was a lot of updating needed due to changes in develop. If so, could you provide a screenshot?

joshcbee:
resolved by #381

PR #381, opened 2019-09-21
Implement view page for lists and correct course_creation_form error …

joshcbee:
…in createList which was leading to crash on attempting to make list with same name and year.

Addresses #365 

View page implemented for lists.

Bug appeared to be a result of uninitialised course_creation_form short-circuiting the error message for uniqueness. All lists must have a unique name, year combination. Expected behaviour on violation is error message to user.

image


JackSekoranja (approved):
Works well, happy for you to merge.

PR #387, opened 2019-09-23
Fix hovering checkboxes by pushing modals above the rest of the page

### LINKED ISSUE #344: Checkboxes Float on Top of Pop Up Windows
When viewing a page with a popup open, if a checkmark appears beneath it then it is both visible and blocks the input of other elements on the popup. Users are able to click the checkboxes even if pop-up windows are on top of them. Sometimes, the checkbox floats on top of the 'Add' button, which prevents users from clicking it.

**Checkmark visible through text box:**
Screen Shot 2019-09-16 at 5 49 31 pm

**Checkmark visible through faded background:**
image

j-selby:
Closes #344.

This PR is a quick fix to fix checkboxes which hover over modals. Due to how checkboxes were implemented, they needed relative positioning with custom z-indexes. Modals also have custom layout logic, but we never specified a z depth which meant that checkboxes could potentially get displayed over them. This PR just specifies a value for this.

fix-checkboxes

### DIFF
--- a/cassdegrees/static/css/style.css
+++ b/cassdegrees/static/css/style.css
@@ -144,6 +144,8 @@ select.text{
     left: 0;
     width: 100%;
     height: 100%;
+    /* Above checkboxes (with a ~5 z-index) */
+    z-index: 10;
 }
 
 .modal-background {


aidengeorge97 (approved):
Simple PR, works exactly as expected.

Issue #374, opened 2019-09-18
Rewording of `non-Public` warning message


JackSekoranja:
In recent UAT, the tester was confused by the warning message that appears when the `Publish` field in programs is not selected (Implemented in PR #351). 

The tester suggested the following wording:
"*This document will not be available to students. Do you want to continue?*"

However, it may be better to specifically mention the `Publish` option as staff will need to know how to allow students to view the plan (i.e. how to fix the warning). The purpose of the PR is to clear this message up a bit.

Issue #345, opened 2019-09-15
Make 'publish' option more obvious to the user


Daniel-Jang:
During UAT, the 'publish' option didn't stand out much so it was missed often. I believe it is a combination of the fact that it is small and that it isn't very descriptive of what it does.

Kind of similar to #321. Perhaps provide a small description of what 'Publish' does and/or make the option stand out more compared to other options.

UAT Link: https://docs.google.com/document/d/1wk9sX77a6Vi7mha6BX23KsZodu_yuyDUunoW9eJvKIs/edit#

KavsHub:
This has also been raised in #316 
Since this is an ongoing issue we should prioritise this.

JackSekoranja:
Closed with PR #388

Issue #316, opened 2019-09-05
Help Tooltip for Publish Checkbox


aidengeorge97:
UAT showed that users may not have a full understanding of what the Publish checkbox does. A tooltip explaining where a published template is visible would be useful for users to have.

PR #388, opened 2019-09-23
Updated messages to make publish clearer to the user

KavsHub:
Addresses issues raised by UAT's constantly; specifically (#374 , #345, #316)

Created a tooltip below the publish checkbox, explaining what it means
image

Updated the wording in the alert notifications
image

I also think we should look into better-looking alert boxes as default ones tend to be ignored as they look like generic popups.

Happy to take suggestions for better wording etc.

Should probably commit this before the UAT on Tuesday...

### DIFF
--- a/cassdegrees/templates/staff/creation/createprogram.html
+++ b/cassdegrees/templates/staff/creation/createprogram.html
@@ -53,6 +53,9 @@
                         {% endif %}
                         {{ field }}
                     </p>
+                    {% if field.name == "publish" %}
+                        <p class="text-center">Please tick the checkbox above as 'published' if you want this to be searchable as a requirement and by students</p>
+                    {% endif %}
                     {% for error in field.errors %}
                          <div class="msg-error inline-error">{{ error }}</div>
                     {% endfor %}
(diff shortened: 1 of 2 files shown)


aidengeorge97 (changes requested):
I just have minor nitpicky suggestions. I think this solution is perfectly reasonable as is but could be polished more.

I think it would be neater to have the publish tooltip only appear on a hover or help button? Or rather than wording it as an instruction have it as a warning that is consistent with the notification "Unpublished programs will not appear to students or be selectable as a requirement elsewhere."

For the popup notification, perhaps this could be formatted to be consistent with our other pop-up dialogs?

j-selby:
For reference, I displayed a confirmation box in one of the earlier related issues because it blocks execution and allows us to fairly simply check stuff before form submission. With that said, this approach could be improved.

KavsHub:
Yep, I'll look into hidden tooltips (may as well make them for other sections while I'm at it)

I also wanted to explore changing the checkbox to maybe a slider or something with draft-publish so that they know that drafts are not searchable

joshcbee:
Will create a new issue, but from today's user testing:
      What do I do with the publish button – correctly surmises that the button would make it live for students – any reason why we have the publish button where it is – seems out of order when first creating a plan – not important in the beginning while still entering the information.
-        Save prompt comes up for publishing again – if I’m saving as I go, I don’t need the publish text, when I save and exit would want to see the publish sign, staying on the page it’s not relevant.

PR #382, opened 2019-09-23
Add verification for plan and subplan vue rules

JackSekoranja:
Closes #380

This PR will add unit verification in program plans and subplans. Furthermore, OR rules will now throw an error if there is an inconsistent number of units in each group. This functionality also supports `At Least`, `Exactly` and `No More Than` rules by bounding the unit count as the plan is evaluated.

**Error Message for Subplan Rules:**
Subplan MSG

**OR Rule Error Message:**
OR MSG

PR #389, opened 2019-09-23
Minimum Maximum Course Rule Option

### LINKED ISSUE #350: Course rule needs to support "At least X but no more than Y" cases
At the end of the UAT with Beth, she pointed out that there are some programs that contains course lists that said "At least X but no more than Y" courses. We currently do not have a way to represent this, so we should add support for such occurrences.

Example program with this: https://programsandcourses.anu.edu.au/2019/program/MLING

UAT Link: https://docs.google.com/document/d/1wk9sX77a6Vi7mha6BX23KsZodu_yuyDUunoW9eJvKIs/edit#

aidengeorge97:
Closes #350.

Enables CASS staff to represent course rules like below in the system.
image

Screenshot:
image

### DIFF
--- a/cassdegrees/static/js/staff/rules.js
+++ b/cassdegrees/static/js/staff/rules.js
@@ -1083,18 +1083,11 @@ Vue.component('rule_either_or', {
                             case "SPEC": units += 24; break;
                         }
                     }
-                    else if (rule.hasOwnProperty("list_type")) {
-                        if (rule.list_type == 'min_max'){
-                            units += parseInt(rule.min_unit_count)
-                        }
-                        else {
-                            units += parseInt(rule.unit_count);
-                        }
-                    }
                     else if (rule.hasOwnProperty("unit_count")) {
                         units += parseInt(rule.unit_count);
                     }
                 }
+
                 if (units > 48) {
                     this.large_unit_count = true;
                     return;
--- a/cassdegrees/static/js/staff/rules.js
+++ b/cassdegrees/static/js/staff/rules.js
@@ -635,6 +635,7 @@ Vue.component('rule_course', {
 
             // Clear options proxy to avoid selection tags from being displayed
             this.optionsProxy = []
+            this.check_options();
         },
 
         // remove the item from the display list and the elements field when x is clicked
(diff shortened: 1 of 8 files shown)


j-selby:
Not directly related, but how would we compute overall units with this approach (i.e to verify that everything adds up at the end)?

JackSekoranja:
> Not directly related, but how would we compute overall units with this approach (i.e to verify that everything adds up at the end)?

I already have a system for this. It basically sums up exact, minimum and maximum units in a dict (`{'exact': 0, 'min': 0, 'max': 0}`) and will accept it if the total unit count falls within this range. Still, there is a potential collision with #382

aidengeorge97:
@Jack, will that system be compatible with what I've built here? I suppose you should be able to test if the course rule is a min_max and if so just add the min_units and max_units instead?

Although I am a little unsure how this rule fits in with an overall plan. I assume the minimum unit count is what is needed and counted towards the total, and the maximum unit count is just an extra constraint

JackSekoranja:
Yeah it’s very easy. You should just be able to modify the ‘count_units’ function to be the same as the other one with Min/Max

aidengeorge97:
Okay Jack, I've added that compatibility in. It counts the min_unit_count from min_max course rules.

aidengeorge97:
I've rebased this to now be compatible with Jack's recent PR which implements unit count verification. I believe this PR is ready to be reviewed again after the discussions.

Daniel-Jang (changes requested):
As long as the values are correct, it works fine. However, when the min is > max, hitting the 'save' or 'save and exit' provides no response, no error messages and stuff. I am happy to approve after this is fixed.

j-selby (changes requested):
This seems to have issues with updating the UI (Firefox 71):

broken-uat-min-max

Adding a new course rule and setting min/max doesn't add the boxes until a redraw occurs.

Code otherwise looks fine!

Daniel-Jang (approved):
Ok turns out the error message only appears in programs. I am happy to approve it now but please ensure it appears for subplans before merging it in to develop.

aidengeorge97:
> Adding a new course rule and setting min/max doesn't add the boxes until a redraw occurs.

I've just tested this with Firefox 69 and 71 and I wasn't able to recreate the issue you described. I'm confident that the dropdown should be causing a redraw because this was an issue I dealt with while developing the feature. Did you do a full ctrl-shift-r refresh? Or are you sure you're running the latest commit?

j-selby (approved):
Bug resolved. Looks good!

Issue #377, opened 2019-09-19
Add functionality to download an Excel sheet of the Course Report


Daniel-Jang:
This extends on issues: #223 #224 

The users may want an option to download the report generated, so provide them the functionality.

PR #390, opened 2019-09-23
Added feature to generate excel files for the courses report.

Daniel-Jang:
Original Issue: #377 

Created a simple button that generates the courses report excel file:
image

Excel output:
image

### DIFF
--- a/cassdegrees/templates/staff/list.html
+++ b/cassdegrees/templates/staff/list.html
@@ -67,8 +67,6 @@
                     {% endif %}
                 </div>
 
-
-
                 {# If there is no data available, notify the user #}
                 {% if not content %}
                     <p class="left fullwidth">No {{ title }}s to display</p>
--- a/cassdegrees/templates/staff/list.html
+++ b/cassdegrees/templates/staff/list.html
@@ -57,7 +57,7 @@
                 {# Allow manage_courses() function to behave appropriately based on origin of request. The post request can either come from list.html or managecourses.html #}
                 <input type="hidden" id="perform_function" name="perform_function" value="retrieve view from selected">
                 <div>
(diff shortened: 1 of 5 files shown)


joshcbee (changes requested):
Looks great and works mostly from what I can test. One small bug, is where there are courses with the same code but different names (e.g. arab6501) the subplan/plan dependency will get added for both where you have only entered one - i.e. add ARAB6501 in person, get:

image

Daniel-Jang:
> Looks great and works mostly from what I can test. One small bug, is where there are courses with the same code but different names (e.g. arab6501) the subplan/plan dependency will get added for both where you have only entered one - i.e. add ARAB6501 in person, get:
> 
> image

The bug you pointed out isn't caused by the code in this branch, but rather, caused by the way we save courses in rules. This bug has been outlined in this issue: #397 

However, it was good for you to point it out, I have not noticed this before.

joshcbee:
Since Aiden's push in #366, courses are stored in rules with code and name.

Still, same question applies as I put on #397 as to whether two courses with the same code should exist or if it's just an idiosyncracy of the list that CASS should remove.

Daniel-Jang:
> Since Aiden's push in #366, courses are stored in rules with code and name.
> 
> Still, same question applies as I put on #397 as to whether two courses with the same code should exist or if it's just an idiosyncracy of the list that CASS should remove.

That's strange, I replicated the issue identified in #397 just yesterday, which includes Aiden's code. Am not sure whats wrong, will investigate.

joshcbee:
Following client meeting - the expectation is that course codes should be unique. Happy to approve.

joshcbee (approved):

PR #391, opened 2019-09-24
Fix custom text rule checkbox rendering

### LINKED ISSUE #361: Newly added rules render the "Old" Checkmarks
When creating a plan, if a new rule is added (e.g. `Custom (Text)`) the old checkmark is rendered instead, possibly because the script to change the checkmarks needs to be re-run. It should be noted that if the page is saved and re-loaded, the checkmark renders correctly.

Screen Shot 2019-09-16 at 6 37 42 pm

j-selby:
Closes #361

This PR fixes the rendering of checkboxes on custom text rules. This was because the JavaScript replacement code didn't detect the checkboxes displayed in Vue's virtual DOM and subsequently displayed the old UI element instead.

screenshot-11_51_26_-_24_09_19

### DIFF
--- a/cassdegrees/static/css/style.css
+++ b/cassdegrees/static/css/style.css
@@ -483,3 +483,9 @@ select.text{
 .pretty .state label {
     overflow: hidden;
 }
+
+/* Forces a checkbox to appear larger on the page */
+.large-checkbox {
+    margin: 5px;
+    font-size: 25px;
+}
--- a/cassdegrees/static/css/style.css
+++ b/cassdegrees/static/css/style.css
@@ -484,6 +484,14 @@ select.text{
     overflow: hidden;
 }
 
+/* A centered checkbox */
+.pretty.centered {
+    display: block;
+    margin-right: auto;
+    width: calc(1em + 2px);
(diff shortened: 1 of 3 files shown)


joshcbee (commented):
Would it be prohibitively hard to have the checkbox inline matching the styling in program details? And is the styling of the new checkboxes in the global requirements meant to be off centre?
If this is as intended though happy to approve.

j-selby:
@joshcbee your image there didn't upload, sorry - is this something that is still a problem?

aidengeorge97:
I think Josh's comment was referring to having the custom text checbox inline with the text, like the publish checkbox is

Have this:
image

Look like this:
image


And in the global requirements, the checkboxes are slightly off center to the left when it might look neater to have them in the center. 
image

j-selby:
I've addressed the global requirements component, but putting the checkbox inline for those 6-unit courses is much more complex as ANU's styling for whatever reason breaks margins there and putting the checkbox inline breaks more stuff than it fixes.

aidengeorge97 (approved):
No worries, I'm happy to approve that given the circumstances.

PR #392, opened 2019-09-24
Uat searchable subplan ui

joshcbee:
Implementation of:
- new list UI for subplans
- addition of optional description fields for lists
- modified input structure of course lists and subplan lists to incorporate the above
- corrections to error checking and when error messages are displayed for new list structure

Note the list changes only apply to the rule creation in programs at this stage as there are discrepancies between rulescripts and subplanrules that need to be resolved for consistency.

image

### DIFF
--- a/cassdegrees/static/js/staff/rules.js
+++ b/cassdegrees/static/js/staff/rules.js
@@ -288,6 +288,8 @@ Vue.component('rule_subplan', {
             "optionsProxy": [],                  // required prop, default behaviour to display selection tags
             "selected_subplans": [],             // used to store the code and name version of the course when selected
             "showLoadingSpinner": false,
+            "subplan_type_label": "",
+            "student_description_label": "",
 
             // Display related warnings if true
             "non_unique_options": false,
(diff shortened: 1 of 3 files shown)


j-selby (changes requested):
This seems to show a "all options must be filled in error" even when there are no obvious missing boxes:

broken-uat-subplan

Reproduction:
- Created an (empty) subplan, only have that subplan in the system.
- Created a new program, set program year, add subplan rule.
- Fill in options
- Expected: Error should disappear

joshcbee:
> This seems to show a "all options must be filled in error" even when there are no obvious missing boxes:
> 
> broken-uat-subplan
> 
> Reproduction:
> 
> * Created an (empty) subplan, only have that subplan in the system.
> * Created a new program, set program year, add subplan rule.
> * Fill in options
> * Expected: Error should disappear

@j-selby can you please confirm you're working from the latest? The UI doesn't appear to match the current version (see screenshots above)

aidengeorge97 (approved):
I didn't have the same issue as James and I agree it looks like he may have been running an older version. Personally I think the new fields format will still cause confusion and the "List name" feels superfluous. There aren't list names on Programs and Courses so CASS staff will probably be at a loss with what to write there.

Granted, I'm happy to accept this for the sake of the refactor as it is functionally sound and the additions are important. UAT will tell us whether things need to be changed further.

PR #398, opened 2019-09-24
Autofill Year for Subplan Created from Program

### LINKED ISSUE #371: Autofill year when clicking "Create New Subplan" in programs
In recent UAT the tester was creating a subplan but entered a year that was different to the one in the plan. As a result, the subplan they created did not appear in the form. To fix this, we should autofill the suplan's year as it should be the same as the program year.

UAT Link: https://docs.google.com/document/d/17vDpVLtRfp05KepaEAJbG118ecldQtxiap_FCheVNnw/edit#

aidengeorge97:
Closes #371.

This simple PR automatically fills the year of a subplan created from the New Subplan button in a program using the year of that program. This negates users accidentally creating subplans with mismatched years that cannot be seen from the program they wanted to add to.

### DIFF
--- a/cassdegrees/templates/staff/creation/createsubplan.html
+++ b/cassdegrees/templates/staff/creation/createsubplan.html
@@ -82,6 +82,10 @@
     {% include "widgets/staff/coursecreatepopup.html" %}
 
     <script>
+        window.addEventListener("load", function(){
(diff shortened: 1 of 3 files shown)


j-selby (approved):
Simple solution, looks good.

Issue #356, opened 2019-09-15
Delete Confirmation Message Has Inconsistent Format/Indentation 


spencerj411:
**Course Deletion**
![image](https://user-images.githubusercontent.com/24206502/64924212-5603f400-d825-11e9-96e0-c208c0e74c09.png)
**Subplan Deletion**
![image](https://user-images.githubusercontent.com/24206502/64924218-6916c400-d825-11e9-9e00-08fc796e241c.png)
**Program Deletion**
![image](https://user-images.githubusercontent.com/24206502/64924222-7a5fd080-d825-11e9-80ea-397f0b63a19c.png)

PR #399, opened 2019-09-24
Made delete messages consistent

KavsHub:
Sorry I forgot about issue #356 

I've made all of the delete page messages consistent now. 
They also change if there are 1 or more for plural cases.

Single
image

Pural
image

Please let me know if there are any issues/nitpicky things.


j-selby (approved):
Looks good.

PR #412, opened 2019-09-26
added subplan type on viewprogram (subplan rule, OR rule)

### LINKED ISSUE #360: Specify Subplan Type When Viewing Programs
When viewing programs, the system will currently say "Completion of one subplan from the following list", whereas it should say "Completion of one Major..." or "Completion of an Arts Major..." (Which was added in the subplan text description).

See an example of the issue below.

Screen Shot 2019-09-16 at 6 29 11 pm

spencerj411:
closes #360 
image


j-selby (approved):
Looks good!

aidengeorge97 (approved):
I agree, this is clean and good to be merged.

PR #411, opened 2019-09-26
Rule and Or Group reordering

Daniel-Jang:
Original Issue: #379 

This Pull Request allows the user to move rules and or group rules up and down. The code is based on James' refactoring branch.

View:
image

Issue #379, opened 2019-09-19
Allow reordering of rule items in programs, subplans and courses


Daniel-Jang:
Currently, there is no known method to reorder the rules. The users expressed that they may want to change the order of the rules.

Nothing fancy is needed, only need 2 buttons (Move up and Move down) for now.

PR #418, opened 2019-09-29
Rule and Or Group reordering Take 2

Daniel-Jang:
Original PR: #411 
Original Issue: #379 

This Pull Request allows the user to move rules and or group rules up and down. The code is based on James' refactoring branch.

View:
image

### DIFF
--- a/cassdegrees/static/js/staff/rules/either_or.js
+++ b/cassdegrees/static/js/staff/rules/either_or.js
@@ -61,6 +61,26 @@ Vue.component('rule_either_or', {
             this.details.either_or[group].push(JSON.parse(JSON.stringify(this.details.either_or[group][index])));
             this.do_redraw();
         },
+        move_up_rule(index, group) {
+            var rules_array = this.details.either_or[group];
+
+            if (index > 0) {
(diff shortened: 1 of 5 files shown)


j-selby (approved):
Everything looks good for this one, but is it possible to hide the move up/down buttons if the item in question is already the top/bottom element respectively?

JackSekoranja:
I still think this should allow users to place rules into/out of OR rules, as that was what the tester was actually referring to when they raised the issue

Daniel-Jang:
> Everything looks good for this one, but is it possible to hide the move up/down buttons if the item in question is already the top/bottom element respectively?

It is possible, but it would require a bit of extra javascript and I'm not sure how to go about with it. Also, I don't think the extra effort required to make it work is worth the gains, in my opinion. Thanks for the suggestion though.

Daniel-Jang:
> I still think this should allow users to place rules into/out of OR rules, as that was what the tester was actually referring to when they raised the issue

Hmm, maybe that should be another issue then? The users also wanted the ability to reorder things, which is what this PR addresses.

Issue #414, opened 2019-09-26
Enforce unique course code


joshcbee:
Following conversation at client meeting, enforce uniqueness of course codes in database.

joshcbee:
Supercedes #397

PR #419, opened 2019-09-29
Unique Course Codes

Daniel-Jang:
Original Issue: #414 

Course codes are now unique and trying to create a new course with same course code results in an error:
image

### DIFF
--- a/cassdegrees/api/models.py
+++ b/cassdegrees/api/models.py
@@ -10,7 +10,7 @@ class SampleModel(models.Model):
 
 class CourseModel(models.Model):
     id = models.AutoField(primary_key=True)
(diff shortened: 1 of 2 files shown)


Daniel-Jang:
Is this error message fine? Or should it be changed to make it more user friendly?

j-selby (approved):
Looks fine, the description naming a model might be a little intuitive.

JackSekoranja (approved):
Personally I'd make a more basic error message (i.e. "A course with this code already exists"). Tested it and it works fine

Daniel-Jang:
Okay, managed to change the error message like the following:
image


Will merge it in now.

PR #420, opened 2019-09-30
Fix caching issues

### LINKED ISSUE #230: Caching results in changes not always being propagated to clients
The current system results in issues where page modifications might not be reflected on client machines, resulting in the need to reset the cache or do a hard refresh.

The system doesn't appear to currently be using mechanisms like E-Tags, which is something which should be investigated going forward, esp. when the admin page changes paths.

JackSekoranja:
Closes #230 

Adds verification to created CSS and JS files to ensure the latest versions are available. This PR will add a version number to each file (set by the last edited time) so that if the server tries to access a CSS/JS file that has been updated, it won't be able to find it in the cache. However, if the file has not changed, it will be able to find it, and will use that file instead. Since you can't send headers when you load CSS or JS, this approach is commonly used.

NOTE: This was only done to our files, so files like `vue.js` will be cached as normal.

### DIFF
--- a/cassdegrees/templates/base.html
+++ b/cassdegrees/templates/base.html
@@ -1,5 +1,6 @@
 {# Follows from https://docs.djangoproject.com/en/2.1/ref/templates/language/ #}
 {% load static %}
+{% load cache_control %}
 <!DOCTYPE html>
 <html lang="en">
 <head>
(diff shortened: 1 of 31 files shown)


j-selby (approved):
Perfect fix - looks good!

PR #423, opened 2019-09-30
Fix course list issues and OR unit count verification

### LINKED ISSUE #417: Updated course rule doesn't call updates when nested in OR rules
OR rules have specific functionality to check whether the user has started to create the rule with more than 48 units. To work correctly, the rule needs to be updated whenever the unit count of its child rules change. The purpose of this PR is to implement this for `Course` rules, as it currently doesn't work

### LINKED ISSUE #422: Fix `course list` rule bugs
The current version of the course list rule has several bugs:
1. `Minimum and Maximum` options don't update the text values
2. Clicking `Minimum and Maximum` and then any other option will sometimes not remove the boxes

Both bugs can be seen here:
Screen Shot 2019-09-30 at 11 33 38 am

JackSekoranja:
Closes #417, Closes #422 

This PR fixes the OR rule verification that didn't work with the courses rule, fixes the `Minimum and Maximum`  option so it updates the text below, and fix an issue where clicking a different option after `Minimum and Maximum` would not update the rule display.

### DIFF
--- a/cassdegrees/static/js/staff/rules/course_selection.js
+++ b/cassdegrees/static/js/staff/rules/course_selection.js
@@ -59,6 +59,8 @@ Vue.component('rule_course_list', {
             "info_msg": INFO_MSGS['course'],
             "list_type_label": "",
             "unit_value_label": "",
+            "max_unit_value_label": "",
+            "min_unit_value_label": "",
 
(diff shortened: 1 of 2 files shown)


KavsHub (approved):
Looks good to me.

PR #394, opened 2019-09-24
Refactor rules

j-selby:
This was painful...

Closes #367

This PR breaks apart rules into individual files, cleaning them up along the way. In doing this, I have attempted to remove duplicated logic, minimise the amount of information loaded for pages (i.e courses doesn't need program-specific rules), reduce LOCs across the board, and get rid of rulescripts entirely.

Further, I have tried to enforce ES6 throughout these new files.

This may break old course lists (due to change in name of vue object) - this was done to reduce ambiguity, but this needs to be verified independently.

PR #424, opened 2019-09-30
added either_or and custom text rules to subplan

### LINKED ISSUE #357: Add `Custom (Text)`  and `either_or` rules in subplan creation
Some subplans have rules that we can't currently express with the current subplan creation system. Consider the Advanced Studies Minor (https://programsandcourses.anu.edu.au/2019/minor/ADST-MIN), especally the rule that is as follows:

> And either:
>>Completion of the equivalent of 12 units in NUS modular credits of Independent Study Modules (ISMs) at National University of Singapore

>Or:
>>12 units from completion of the following course(s):
>>...

spencerj411:
closes #357 
Thanks to James' #394, it made this task easy to complete.
image
**Edit:**
Custom Text Rule in Subplan:
image
**Edit:**
This PR also includes the fix for #451 .


JackSekoranja (changes requested):
Functionally it works well. There are still some elements that don't make sense for subplans that we use in programs (e.g. `Show 6-unit course boxes in plans (e.g. PDF):`). It would be great to remove this before merging.

Also this is not a result of this PR, but it look like options for subplans and elective lists appear in the subplan rule list, so we might need to create a new issue around that.

JackSekoranja (changes requested):
The `Custom (Text)` rule is blank to me, even after refreshing my cache. Also I realised the other rules being blank is the result of this PR, so it would be great to only include the three relevant ones. This could be done by specifying a new dict in `rules.js` for relevant subplan rules, and then adding only the relevant list in `subplan_rules.html`

j-selby (approved):
Looks good, works fine!

PR #425, opened 2019-10-01
Modify PDF styling to improve various aspects of display

### LINKED ISSUE #385: Improve PDF Rendering
There are a few issues with PDF rendering that we need to iron out as follows:

1. When rendering OR rules, both options are rendered where only spaces for one should be. (Part of this task will be determining a suitable compromise)
2. Some box headings can appear in one column, while the content can appear in another. If possible, it would be good to change the appearance of elements based on their size so we can fill more of the screen.
3. When compulsory courses are added to the PDF, it appears that they have just been filled in, whereas on the current CASS plans they have checkboxes for students. Not saying we need to replicate checkboxes but it is something to be considered.
4. "Instructions for filling in" always appears on next page (I know it was intentional at the time but it isn't necessary)
5. Generally cleaning up the templates themselves. Make it look as good as you can and add any other features you think may be nice

Consider the following plan for replication ([BMUSI 2019](https://programsandcourses.anu.edu.au/2019/program/BMUSI)):
https://cass.anu.edu.au/sites/default/files/BMUSI%202019.pdf

### LINKED ISSUE #416: Student-facing plan creation broken
Currently the student-facing side cannot load any new or old plans, showing a `string indices must be integers` error. This basically breaks the entire student-facing system.

### LINKED ISSUE #440: Bug when rendering plans on the student-side
Original UAT: https://docs.google.com/document/d/1eOsRcQOtWZSDc-dio8-Mni9_g6zNw3X5m_AzjxYiqk0/edit#heading=h.fgpu8pdjhtwe

App crashes when you try to create a plan based off on an existing program on the student-side.

It should be simple to fix, so I will label as easy for now.

### LINKED ISSUE #442: PDF pushes components into the next page when unnecessary
Original UAT:
https://docs.google.com/document/d/1eOsRcQOtWZSDc-dio8-Mni9_g6zNw3X5m_AzjxYiqk0/edit#

The PDF view pushes components into the next page, even when there is plenty of space in the first page, leaving big empty spaces.

j-selby:
Closes #385.

Closes #442.

Should close #440.

(Another big one, sorry!)

This PR fixes a few issues with PDF styling to improve both the overall layout and to improve fonts/consistency.

Firstly, this fixes issues with OR rules rendering multiple course boxes for all the branches. This now renders this as one. This does remove any pre-filled values from the template, though these are now listed as part of the rule description itself when this occurs.

Student filled values will still fill as per normal (note the changes as well to the student-facing frontend to fix that there as well).

The top title has been tweaked to maximise space, noting that if a student put in a really long title this would cause issues. This provides (at least) a little more space there.

Breaks between boxes has been removed, noting that a smaller plan was preferred in testing (ideally, one page). Further, we now render 4 columns (rather than 3) as this doesn't really affect readability and allows us to get more course boxes onto the page. Page margins have also been tweaked.

Edit: Sample image:
DaringIssue

**Jack's Edit:** Will also Close #416, as through testing this PR seems to remove the bug

### DIFF
--- a/cassdegrees/templates/pdf_base.html
+++ b/cassdegrees/templates/pdf_base.html
@@ -47,6 +47,10 @@
             padding: 0.25cm;
         }
 
+        .sub-box {
+            padding-bottom: 0.1cm;
+        }
+
         .box p {
             margin: 0;
         }
--- a/cassdegrees/templates/pdf_base.html
+++ b/cassdegrees/templates/pdf_base.html
@@ -91,7 +91,7 @@
         }
 
         .try-complete-box {
-            page-break-inside: avoid;
+            break-inside: avoid-page;
         }
     </style>
 
--- a/cassdegrees/templates/pdf_base.html
+++ b/cassdegrees/templates/pdf_base.html
@@ -18,7 +18,7 @@
 
         @page {
             size: A4 landscape;
-            margin: 1cm;
+            margin: 0.5cm;
         }
 
         /* Basic column rules */
@@ -29,6 +29,11 @@
 
         .columns-3 {
             column-count: 3;
+            column-fill: auto;
+        }
+
+        .columns-4 {
+            column-count: 4;
             column-fill: auto;
         }
 
@@ -53,6 +58,11 @@
         .no-bottom-margin {
             margin-bottom: 0 !important;
             padding-bottom: 0 !important;
+        }
+
+        .no-top-margin {
+            margin-top: 0 !important;
+            padding-top: 0 !important;
         }
 
         .grey-text {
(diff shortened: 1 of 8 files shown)


JackSekoranja (changes requested):
Pretty good, but in UAT we realised that longer rules will simply be placed on the next page even when there is space on the first page. It would be great if we could sort out this issue first:

UAT Link: https://docs.google.com/document/d/1eOsRcQOtWZSDc-dio8-Mni9_g6zNw3X5m_AzjxYiqk0

Screen Shot 2019-10-03 at 11 11 59 am

JackSekoranja (approved):

PR #429, opened 2019-10-01
Added the ability to clear a recent search.

### LINKED ISSUE #404: Navigation from search page
After entering a search from /staff/list/, there is no way to get back to the generic /staff/list/ page without clicking back on the browser button, suggest:
- Making the 'Program, Subplan, and Course List' heading clickable to get back; and/or 
- adding an additional breadcrumb for the step back.

image

aidengeorge97:
Closes #404.

I chose to solve this problem by adding a "Clear Search" button which should act as a clear way to return to the unfiltered list. 

image

### DIFF
--- a/cassdegrees/static/css/style.css
+++ b/cassdegrees/static/css/style.css
@@ -43,9 +43,9 @@
 
 .custom_search_bar input[type=text]{
     height: 30px;
-    width: 78%;
+    width: 67%;
     padding-left: 5px;
-    margin-right: 2%;
+    margin-right: 1%;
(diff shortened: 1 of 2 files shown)


JackSekoranja (approved):
Nice and simple. Would be great of the "Clear" option was provided as a little cross on the right side of the text field, but this works too

j-selby (changes requested):
The clear button here doesn't appear correctly at smaller resolutions:
broken-clear

Works well otherwise!

PR #430, opened 2019-10-01
Custom no courses error and changes to rendering order.

### LINKED ISSUE #393: Attempting to create an empty list results in unintuitive error
When attempting to create a course list, filling in fields but not adding any courses, an error is shown when attempting to save this list which doesn't demonstrate well that the issue with this is related to the lack of courses, rather than the year or otherwise. This would likely require some kind of manual verification before checking.

image

joshcbee:
Resolves #393 

image


JackSekoranja (approved):
Works well. Happy for you to merge

PR #431, opened 2019-10-02
Implement Several Small Improvements

JackSekoranja:
This PR implements the following changes in preparation for todays UAT:

- **Elective Rule:**
   - Default value is now 24 instead of 0
   - Values of 0 are now considered invalid (Can't complete 0 units of electives)
   - Minimum value of the unit scroll option is now 6 instead of 0
- **Course Selection Rule:**
   - Removed debug print statement
   - When choosing from the `Minimum and Maximum` option, a minimum of 0 is now valid
   - Error messages now disappear as soon as they are fixed
   - Standard and Max unit count boxes now have the minimum set to 6 instead of 0
- **Custom (Text) Rule:**
   - Default value now set to 6
   - Added verification on 
   - Minimum value of the unit scroll option is now 6 instead of 0
   - Step value was set to 6 for unit count

### DIFF
--- a/cassdegrees/static/js/staff/rules/course_elective.js
+++ b/cassdegrees/static/js/staff/rules/course_elective.js
@@ -6,7 +6,7 @@ Vue.component('rule_elective', {
             validator(value) {
                 // Ensure that the object has all the attributes we need
                 if (!value.hasOwnProperty("unit_count")) {
-                    value.unit_count = 0;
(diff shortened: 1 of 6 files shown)


aidengeorge97 (approved):
Good quality of life changes. Nice and simple.

Daniel-Jang (approved):
Seems good!

PR #434, opened 2019-10-02
Select All (In Current Page/All Entries) in Course/Subplan/Program List

### LINKED ISSUE #276: Pagination results in the "Select All" to only select all courses on that page; no option to mass delete everything
Is this a desirable feature? Just selecting the courses on the current page?
If yes, should we add an option to delete everything?

spencerj411:
closes #276.

Note: Inspired by how Gmail does Select All.
What happens is:
- If you click on the 'Select All' checkbox, all the current checkboxes on the screen will be checked. A popup message will then be shown:
image
- If you click on 'Select all Courses/Subplans/Programs' in the popup message, the pagination will be updated and all entries will be shown on screen. All the checkboxes on the screen will be checked. The message on the popup will update and tell you that you have selected all entries:
image
- As this implementation relies on the entries being shown on the screen, I have moved the 'Delete Selected' button from the bottom to the top, next to the 'Add' button. Also, I implemented the most basic form of sticky header (the buttons will stay at the top when you scroll down):
image
I'm not too sure of the idea of putting all the entries on the page when wanting to select all entries, but it's the only way I could do it where all the checkboxes can be seen as checked (even for gmail, when you select all entries, and you go through the pages, only the first page's checkboxes are checked).

Another Note: this took me way longer than expected due to not being experienced with javascript. Fortunately, I was able to eventually create a solution.

### DIFF
--- a/cassdegrees/static/css/style.css
+++ b/cassdegrees/static/css/style.css
@@ -41,6 +41,13 @@
     width: 98.36%; /* textarea need custom width as anu styling width makes it longer */
 }
 
+.list_menu {
+    height: 50px;
+    background-color: white;
+    position: sticky;
+    top: 0px;
+}
+
 .custom_search_bar input[type=text]{
     height: 30px;
     width: 78%;
--- a/cassdegrees/static/css/style.css
+++ b/cassdegrees/static/css/style.css
@@ -48,6 +48,10 @@
     top: 0px;
 }
 
+.shadow {
+    box-shadow: 0 5px 0 rgba(0,0,0,0.1);
+}
+
 .custom_search_bar input[type=text]{
     height: 30px;
     width: 78%;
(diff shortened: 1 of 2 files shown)


KavsHub (approved):
Very nice feature and works as expected.

One thing you could check is when there's only 1 page of items it says 'All xxx are selected." instead of having the option to select all as its redundant.
However, this case is rare as in production there will always be more than 1 page so probs not worth the effort.

**Also note for the team, should we do some sanity checking on the delete function? Someone can literally delete everything if they select all and press delete.**

JackSekoranja (approved):
Works pretty well. I noticed that when you select all and then unselect, the single large page of results isn't removed. I also think it would be better if the delete button was right-aligned so that it wasn't as close the the create buttons. Both of these are pretty minor issues though.

PR #445, opened 2019-10-03
View/Download Bulk Upload Failures

### LINKED ISSUE #401: Bulk upload - can't see what has failed
Simplification of the error message to "... and 141 more items" when uploads fail mean users can't tell what has failed. Feedback from user:
- I don't care about the list of successes - the number is enough
- For failures, I want to be able to see everything that has failed so I can correct or check it, either through expandable 'more' (what they expected would happen), by a link to a new page with the list or a link to download a csv of failures.

aidengeorge97:
Closes #401.

Added a link where users can view ALL the courses/subplans that failed to be uploaded, and can download them as a CSV file. 

Removed the listings for successfully uploaded items and just replaced it with a number of successes. 

image

image

### DIFF
--- a/cassdegrees/static/css/style.css
+++ b/cassdegrees/static/css/style.css
@@ -156,6 +156,11 @@ select.text{
     background: rgba(0, 0, 0, 0.5);
 }
 
+.modal-body {
+    max-height: calc(100vh - 210px);
+    overflow-y: auto;
+}
+
 .modal-card {
     position: absolute;
     left: 35%;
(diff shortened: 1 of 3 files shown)


joshcbee (approved):
Works well, csv looks good.

JackSekoranja (approved):
Very nicely done. CSV is a nice touch. Only issue I found was that the link to bring up the failed courses doesn't change the cursor to the `click` hand. Otherwise looks good

Issue #355, opened 2019-09-15
Add Delete Feature in Edit List


spencerj411:
Once #290 is done, add the delete feature in _edit/list_, same as #289.

PR #453, opened 2019-10-06
Delete feature in Edit List

Daniel-Jang:
Original issue: #355 

This one is simple - will merge it into develop after 1 approval.

Essentially, there is now a delete button when you are editing lists:
image

Clicking on it:
image


j-selby (approved):
Looks good, works fine.

PR #456, opened 2019-10-07
Truncate courses in course lists in PDFs to avoid filling up entire page

### LINKED ISSUE #349: Fix potentially extremely long course lists
At the end of the last UAT, Beth mentioned that there are some programs with really long course lists (with "no more than Y courses"), which may be problematic with our PDF generation.

An example provided by Beth was the following:
https://programsandcourses.anu.edu.au/2019/program/MLING

Note that Deborah did not prioritise Masters, so I will label this as a 'Wishlist'.

UAT Link: https://docs.google.com/document/d/1wk9sX77a6Vi7mha6BX23KsZodu_yuyDUunoW9eJvKIs/edit#

j-selby:
Closes #349

This PR truncates courses in course lists to prevent issues where a large course list can consume way too much space. It would be much more preferable for students to consult P&C in these cases as it isn't feasible to communicate all possible course combinations in templates.

Preview:
VacantCloud

### DIFF
--- a/cassdegrees/templates/pdf_base.html
+++ b/cassdegrees/templates/pdf_base.html
@@ -58,6 +58,10 @@
             break-after: page;
         }
 
+        .break-page {
+            break-after: left;
+        }
+
         .no-bottom-margin {
             margin-bottom: 0 !important;
             padding-bottom: 0 !important;
--- a/cassdegrees/templates/pdf_program.html
+++ b/cassdegrees/templates/pdf_program.html
@@ -21,7 +21,7 @@
     <div class="columns-4 columns-no-gap">
 
         {% for rule in program.rules %}
-            {% include "widgets/pdf_rules.html" with render_rules=True render_units=True %}
+            {% include "widgets/pdf_rules.html" with render_rules=True render_units=True cache_large_lists=True %}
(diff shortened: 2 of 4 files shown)


joshcbee:
What sort of difficulty would we be looking at to have the lists print on a second page with a list title, then refer to, e.g. Complete 6 units from Methods Courses 1. 

From a usage point of view, I think we run into the same issue we ran into with deleted courses with the 'and 5 more'. If I don't have a way to readily check what those are, the list as a whole is of no use unless I have done exactly those 5 courses listed, and it doesn't help me with where to go next.

What do you think James?

j-selby:
@joshcbee I'll take a look at outputting course lists onto another page. The argument here was that they are going to need to look at programs and courses at some point anyway - the PDF tool has always been to help students make sure they are ticking boxes, not to replace what is available on P&C (as it isn't realistically possible to encapsulate all information).

JackSekoranja (approved):
I find it odd how it prints like this:

```
Complete 48 units from a 
selection of:
See appendix #1 for
complete course list. 
```

It still works though, so I'm happy for you to merge if you don't mind the wording.

PR #457, opened 2019-10-07
Let programs with different types share the same course code

### LINKED ISSUE #443: Tag Degrees with the plan type and make it unique
Original UAT: https://docs.google.com/document/d/1eOsRcQOtWZSDc-dio8-Mni9_g6zNw3X5m_AzjxYiqk0/edit#

BARTS can both be single or double flexible (same code), but perhaps differentiate by adding tags like such: (i.e. BARTS Single Pass and BARTS Flexible-Double-Degree)

j-selby:
**Migration needed!**

Closes #443

This PR changes programs such that they are only unique in combination with their program type, allowing degrees which share codes to have the same course code.

This PR also cleans up constraint checking in forms to be more consistent and cleaner for all usecases in that particular file.

### DIFF
--- a/cassdegrees/api/models.py
+++ b/cassdegrees/api/models.py
@@ -97,4 +97,4 @@ class ProgramModel(models.Model):
     rules = psql.JSONField(default=list)
 
     class Meta:
-        unique_together = (("code", "year"), ("name", "year"))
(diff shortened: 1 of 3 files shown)


joshcbee (approved):
Works well across programs and subplans for constraint checking. Nice work on the refactor. Happy to approve.

PR #459, opened 2019-10-07
Implement check to verify that unsaved changes are acknowledged

### LINKED ISSUE #408: Wishlist - warning when leaving the page with unsaved information
When a user navigates away from the page, either through browser back button to clicking a link, it would be nice to get a warning if there is unsaved information before it is lost.

j-selby:
Closes #408

This PR implements a check to ensure that users aren't losing work when they navigate off a page after making changes to any field. This won't occur in scenarios where a user is submitting a form, but will based on browser navigation or clicking on links. Note that this will not work in all browsers - a few niche browsers intentionally block the ability to block page navigation.

This wasn't been implemented on the student facing side as that uses a different blob of js and would need further investigation.

In Firefox, for example:

change-page

### DIFF
--- a/cassdegrees/static/js/main.js
+++ b/cassdegrees/static/js/main.js
@@ -80,3 +80,69 @@ function updateCheckboxes() {
 }
 
 document.addEventListener("DOMContentLoaded", updateCheckboxes);
+
+// Try to catch instances where users might enter text
+let pageDirtied = false;
+let contentsSubmission = false;
+
+let appContents;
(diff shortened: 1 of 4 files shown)


KavsHub (approved):
Looks good, and works as expected. (Tested on chrome with extensions)
Didn't notice anything out of the ordinary.

Daniel-Jang (commented):
Works well, approved.

Daniel-Jang (approved):

PR #460, opened 2019-10-07
Fix outputting of JSON on student page for specific rules

j-selby:
1-liner to fix issue with student frontend where certain rules would output JSON instead of course codes.

Previous:
student-json


aidengeorge97 (approved):
Looks good!

PR #461, opened 2019-10-07
Draggable rules

JackSekoranja:
After a lot of playing around I finally have this working. Currently this lets you drag both or groups and general rules. You can also drag rules in to OR rules, drag groups between different or rules, etc. 

A lot of this was kind of hackish due to the communication between different elements that don't like talking to each other (Vue, Interact, Javascript DOM, etc.). I tried a number of other solutions for each but couldn't do anything else for some. Still, if you think something could be done better let me know.

Screen Shot 2019-10-07 at 6 23 18 pm
Screen Shot 2019-10-06 at 11 38 34 pm

EDIT: 
Dragged rules now shrink and remove their previous space. 
Screen Shot 2019-10-08 at 11 20 26 pm

### DIFF
--- a/cassdegrees/static/css/style.css
+++ b/cassdegrees/static/css/style.css
@@ -120,8 +120,13 @@
 
 .hovering {
     z-index:100;
-    box-shadow: 5px 10px 8px #888888;
+    box-shadow: 15px 30px 24px #888888;
     position: relative;
+    opacity: 0.8;
+}
+
+.hidden-outer {
+    height: 0px;
 }
 
 
--- a/cassdegrees/static/css/style.css
+++ b/cassdegrees/static/css/style.css
@@ -107,6 +107,20 @@
     color:#FFF;
 }
 
+.dropzone-area {
+    width: 100%;
+    height: 60px;
+    border: 2px green dashed;
+    margin-bottom: 10px;
+}
+
+.hovering {
+    z-index:100;
+    box-shadow: 5px 10px 8px #888888;
+    position: relative;
+}
+
+
 /* Adds carets to list sorting options */
 table th {}
 .sort:after{
--- a/cassdegrees/static/css/style.css
+++ b/cassdegrees/static/css/style.css
@@ -107,6 +107,10 @@
     color:#FFF;
 }
 
+.draggable-group {
+    background-color: white;
+}
+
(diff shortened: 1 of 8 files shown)


aidengeorge97 (changes requested):
This is a really nifty PR, but I did notice a handful of graphical issues.

First, the new rule pop-up dialog gets isolated to an either or rule, when it should cover the full screen.
image

Steps to reproduce:
* Create an either or rule
* Drag the rule and place it back in its original location
* Attempt to create a new rule inside the either ors rule group

Another (very minor) graphical issue I noticed was that either or rules seem to generate two placement options where there should be only one? 
image

Notably, this issue doesn't appear when all of the rule groups are removed from the either or.
image

Here I attempted to drag an elective rule from a rule group in one either or, into a rule group in a separate either or. I don't think the dragged item is supposed to be underneath a place rule.
image

I had another issue where a rule I dragged from one either or rule group into another either or rule group did not appear immediately. The rule did appear after the group it was in was updated.
Steps to recreate:
* Create an either or rule
* Create a second either or rule
* Create a course list rule in the first rule group of the first either or rule
* Drag the course list rule into the first rule group of the second either or rule
* The course list rule will disappear.
* Attempt to create a new rule in the second either or rule.
* The course list rule will reappear where it should be.

I also wonder if there's a way to make it clear to users that these boxes are now draggable? The cursor changing to a draggable cursor is good, but we could do more. Perhaps when hovering over a draggable area, show a little tool tip? Or include some text above all the rules like "Click and drag rules to rearrange them."

Otherwise this functionality is great and works fine, I just noticed those cosmetic issues.

j-selby (approved):
Nicely done and looks really good!

I did notice that after dragging elements that because page layout changes considerably its easy to lose your place on the page, especially when dragging around page-long either/or rules. Is it possible to minimise elements that are being dragged around (i.e only the title bar) and/or to focus on the dragged element upon completion (it doesn't appear that this is the case right now). Not a biggie, but something to note.

aidengeorge97 (approved):
The recent changes have solved the bugs I noted and generally improved the quality of this feature. I'm happy to approve.

Daniel-Jang (approved):
Looks really good, tried it out and I am very impressed. Approved.

PR #463, opened 2019-10-08
Remove Flexible Vertical Option from Program Types

### LINKED ISSUE #444: Remove flexible vertical options from plan types
Original UAT: https://docs.google.com/document/d/1eOsRcQOtWZSDc-dio8-Mni9_g6zNw3X5m_AzjxYiqk0/edit#

Since vertical options don't change any aspect of the program, there isn't any need to be able to distinguish a program as 'flexible vertical'. The second flexible program is just simply appended to the first one.

aidengeorge97:
Closes #444.

Simple one liner to remove an unnecessary option. Requires migrations.

### DIFF
--- a/cassdegrees/api/models.py
+++ b/cassdegrees/api/models.py
@@ -88,7 +88,6 @@ class ProgramModel(models.Model):
                      ("mast-sing", "Masters Single Degree"),
                      ("mast-adv", "Masters (Advanced) Degree"),
                      ("mast-doub", "Masters Flexible Double Degree"),
-                     ("vert-doub", "Vertical Flexible Double Degree"),
                      ("other", "Other Degree"))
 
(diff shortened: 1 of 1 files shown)


KavsHub (approved):
This is fine. 
However, are we no longer doing vertical option?
If so I'll remove it from the student planner and portal too

Daniel-Jang (approved):
Looks fine to me, approved.

aidengeorge97:
> This is fine.
> However, are we no longer doing vertical option?
> If so I'll remove it from the student planner and portal too

From the UAT it seems like a vertical flexible double is just two programs stacked on each other so there's no need to represent it specially. Yea, it might be unnecessary to have it on the student planner/portal  too

PR #465, opened 2019-10-08
Tooltips on Add Rule Popup Dialog

### LINKED ISSUE #317: Display Explanatory Tooltips for Rules on the New Rule Pop-up Dialog
Users who are unfamiliar with the purpose/use of each rule when creating templates appear to be confused when choosing a new rule. Tool tips for rules are only visible after a rule is created. It would be good to show explanations for rules in the Add Rule pop-up dialog so users can make an informed decision about which rule to choose.

aidengeorge97:
Closes #317.

Adds explanations for rules to the pop up dialog so that users can more quickly see and understand what the purpose of each rule is. 

image


Daniel-Jang (approved):
I like it, happy to approve.

Issue #427, opened 2019-10-01
Issue with the .xls file generated by Course Report site


Daniel-Jang:
There is an issue where the .xls file generated by the course report site causes the bulk uploader to crash with error: "BadZipFile: File is not a zip file" when the file is uploaded and submitted. 

I had a preliminary look and it appears that the excel sheet generator in the course report site isn't creating a normal excel sheet and may be missing a bunch of metadata found in normal excel sheets.

There may be a possibility that a new excel writing library has to be used to fix this issue.

Not sure of the difficulty, so I have set it to a medium for now.

PR #470, opened 2019-10-09
Fix Course Report from generating malformed excel files

Daniel-Jang:
Original Issue: #427 

Used openpyxl instead of xlwt to generate excel files, as it turns out xlwt did not generate a well formed excel file. Now, with a well formed excel file generated, it works with bulk uploader well.

Example output from Course Report (try this in bulk uploader):
[Course Report.xlsx](https://github.com/cass-degrees/CASS-Degrees-Code/files/3705737/Course.Report.xlsx)

Also changed bulk uploader excel reading function to open the very first worksheet, instead of trying to find 'Sheet1' specifically and crashing if not found.


joshcbee (approved):
Works well, generates the full course list as an excel file quickly, file reuploads without error, also works with multiple sheets added (selecting first sheet). Happy to approve.

PR #471, opened 2019-10-09
Publish Button Prompt and Workflow

### LINKED ISSUE #407: Location of publish button and publish prompt
Optional improvement: location of the publish button is counterintuitive, as I am just putting in the details for creation of the plan but am being prompted to publish, would expect this at the end/elsewhere.

Workflow change: users should not see the publish popup prompt every time they save as they are likely to save the page progressively as they go for reassurance (and given lack of autosave). They should only see the prompt if clicking save and exit. If saving and remaining on page, prompt should not appear.

image

image

aidengeorge97:
Closes #407.

This PR moves the publish button from the top of the workflow to the bottom, where it was suggested it would be more useful. It also clarifies what publishing an item does.

Additionally, users will only be prompted that their program is unpublished when saving and exiting, not when just saving. 

I also fixed a minor issue where the delete buttons still showed rubbish bins on the program creation pages, whereas it should be a little 'x' like on the list page.

image

### DIFF
--- a/cassdegrees/templates/staff/creation/createcourse.html
+++ b/cassdegrees/templates/staff/creation/createcourse.html
@@ -31,7 +31,7 @@
             {% csrf_token %}
             <input type="hidden" name="id" value="{{ request.GET.id }}">
             <div class="fullwidth auto-overflow">
-                <input class="btn-uni-grad btn-small right" type ="submit" formaction="/staff/delete/courses/" value="&#128465Delete">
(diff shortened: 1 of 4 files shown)


joshcbee (approved):
Works well in both programs and subplans. Happy to approve.

JackSekoranja (approved):
Looks good and works well, happy for you to merge

PR #473, opened 2019-10-09
Bugfix   duplicate error on course edit

### LINKED ISSUE #466: Duplicate Error Prevents Users from Editing Courses
There exists an error where users cannot edit an existing course because it will register as being a duplicate, despite being unique in the system.

Attempting to edit and save this course results in an error, when no error should occur.
image

joshcbee:
Raise validation error if user attempts to make a course with the same course code as a course in the database, where the database IDs for the form course and the database course do not match. Closes #466 

Expected behaviour:
- new course with same code - validation error
- edit existing course - success
- edit existing course with different code to have same code - validation error

Editing existing course:
image

Adding new course with same ID:
image


JackSekoranja (approved):
Simple and works well. Happy for you to merge

PR #480, opened 2019-10-10
Add course list rule to `View` OR rules

### LINKED ISSUE #479: Course lists don't appear in `view` page
When creating program plans, any  `Course List` rules don't appear at all when viewing the program in `http://127.0.0.1:8000/staff/view/program/?id=...`. To close this the rules need to appear correctly

**Edit:** This issue is only present when the course list is in an `Either Or` rule

JackSekoranja:
Closes #479 

*As a bonus, this PR also prevents the user from being able to drag child components of OR groups - only the group border themselves should be draggable.*

Previously the view logic for the `Course List` rule wasn't implemented inside OR rules. This PR makes it so course lists now appear when viewing programs again

### DIFF
--- a/cassdegrees/static/css/style.css
+++ b/cassdegrees/static/css/style.css
@@ -363,6 +363,10 @@ select.text{
     user-select: none;
     cursor: move;
 }
+/* Reverts the draggable CSS changes */
+.non-draggable {
+    cursor: default;
+}
 
 /* Ensures course drop boxes are of the right size */
 .course-drop {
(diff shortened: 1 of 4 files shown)


aidengeorge97 (approved):
I added an additional small fix to prevent elective or rules from appearing as invalid when viewing programs.

Happy for this to be merged.

KavsHub (approved):
Looks good to me. Didn't find any issues.

PR #481, opened 2019-10-11
Fix deleting courses in lists from returning no error

### LINKED ISSUE #454: Course deletion doesn't check if courses are used in lists (Enquiry?)
Courses that are used in subplans or programs can't be deleted, but courses used in lists can still be deleted without any warning or failure. Is this an intentional design?

Example:
List with courses:
image

Courses page after attempting to delete all courses:
image

j-selby:
Closes #454.

Preview:
course-delete

### DIFF
--- a/cassdegrees/templates/staff/creation/createlist.html
+++ b/cassdegrees/templates/staff/creation/createlist.html
@@ -101,10 +101,16 @@
 
     <script>
         function submit_course_form(){
+            // Disable check for unsaved changes - we are saving them here!
+            contentsSubmission = true;
+
             document.getElementById('mainForm').submit();
         }
 
         function submit_form(form_action, redirect) {
+            // Disable check for unsaved changes - we are saving them here!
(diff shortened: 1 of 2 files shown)


spencerj411 (changes requested):
When I create a course list within creating/editing a Program, and then I try to delete a subset of those courses after, an error pops up; looks good.

I'm still unsure about the case of creating/editing a List, and then when I try to delete a subset of those courses after, the deletion goes through. Is this intentional? Thanks.

j-selby:
@spencerj411 I was unable to replicate your second issue:
- I created a list
- Added a course to it, saved it
- Attempted to delete that course, which failed

Do you have specific replication steps?

spencerj411:
nvm, both cases look good.

spencerj411 (approved):

JackSekoranja (approved):
Worked for me, happy for you to merge

PR #482, opened 2019-10-11
Validation After Submission

### LINKED ISSUE #287: Error Message/Validator Popping Up Too Early
When adding a new rule/global requirement to a program, the error message appears instantly. Please make it so the messages only appear when the user is trying to save or when they do an action that's unrelated to making that specific rule/global requirement.

### LINKED ISSUE #436: Don't show global requirement issues until after submitting
This issue was raised in this weeks UAT. Basically the Global Requirements rules should not throw errors until after users have tried to submit once. Almost all other rules in the program rues have this same functionality, so we just need to copy that.

UAT Link: https://docs.google.com/document/d/1eOsRcQOtWZSDc-dio8-Mni9_g6zNw3X5m_AzjxYiqk0

spencerj411:
closes #436 and closes #287 .

We had a few problems where the following will show their validation messages straight after creating the rule/requirement/requisite/incompatibility:
- Global Requirement (in Program and Subplan)
- Custom Text Rule (in Program and Subplan)
- Incompatibility, Enrolled Program, Course Requisite, Custom Text Requisite (in Course)

This PR attempts to solve this.
The validation message will only show if the user attempts to submit.

### DIFF
--- a/cassdegrees/static/js/staff/globalrequirements.js
+++ b/cassdegrees/static/js/staff/globalrequirements.js
@@ -83,23 +83,25 @@ Vue.component('global_requirement_general', {
         const request = new XMLHttpRequest();
 
         request.addEventListener("load", () => {
-            rule.check_options();
+            rule.check_options(false);
         });
         request.open("GET", "/api/search/?select=code&from=course");
         request.send();
     },
     methods: {
-        check_options() {
-            this.invalid_units = this.details.unit_count <= 0;
-            this.invalid_units_step = this.details.unit_count % 6 !== 0;
-            this.units_is_blank = this.details.unit_count === "";
+        check_options(is_submission) {
+            if (is_submission) {
+                this.invalid_units = this.details.unit_count <= 0;
+                this.invalid_units_step = this.details.unit_count % 6 !== 0;
+                this.units_is_blank = this.details.unit_count === "";
 
-            this.is_invalid = !this.details.courses1000Level && !this.details.courses2000Level && !this.details.courses3000Level
-                && !this.details.courses4000Level && !this.details.courses5000Level && !this.details.courses6000Level
-                && !this.details.courses7000Level && !this.details.courses8000Level && !this.details.courses9000Level
-                && this.details.subject_area === "";
+                this.is_invalid = !this.details.courses1000Level && !this.details.courses2000Level && !this.details.courses3000Level
+                    && !this.details.courses4000Level && !this.details.courses5000Level && !this.details.courses6000Level
+                    && !this.details.courses7000Level && !this.details.courses8000Level && !this.details.courses9000Level
+                    && this.details.subject_area === "";
 
-            return !this.is_invalid && !this.invalid_units && !this.invalid_units_step && !this.units_is_blank;
+                return !this.is_invalid && !this.invalid_units && !this.invalid_units_step && !this.units_is_blank;
+            }
         }
     },
     template: '#generalGlobalRequirementTemplate'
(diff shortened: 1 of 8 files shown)


cassdegrees/static/js/staff/globalrequirements.js
```diff
@@ -83,23 +83,25 @@ Vue.component('global_requirement_general', {
...
+                this.invalid_units_step = this.details.unit_count % 6 !== 0;
+                this.units_is_blank = this.details.unit_count === "";
+
+                this.is_invalid = !this.details.courses1000Level && !this.details.courses2000Level && !this.details.courses3000Level
+                    && !this.details.courses4000Level && !this.details.courses5000Level && !this.details.courses6000Level
+                    && !this.details.courses7000Level && !this.details.courses8000Level && !this.details.courses9000Level
+                    && this.details.subject_area === "";
+
+                return !this.is_invalid && !this.invalid_units && !this.invalid_units_step && !this.units_is_blank;
+            }
```
j-selby, line 117: What is the return value if this doesn't return? I would define that

j-selby (changes requested):
Seems fine to me! Just a few things:

cassdegrees/static/js/staff/globalrequirements.js
spencerj411: my bad, thanks james.

cassdegrees/static/js/staff/globalrequirements.js
```diff
@@ -83,22 +83,23 @@ Vue.component('global_requirement_general', {
...
         });
         request.open("GET", "/api/search/?select=code&from=course");
         request.send();
     },
     methods: {
-        check_options() {
+        check_options(is_submission) {
             this.invalid_units = this.details.unit_count <= 0;
             this.invalid_units_step = this.details.unit_count % 6 !== 0;
             this.units_is_blank = this.details.unit_count === "";
```
JackSekoranja: These are set whether or not `is_submission` is true, meaning errors can be introduced at any time. This goes against what the PR is saying it accomplishes

cassdegrees/static/js/staff/globalrequirements.js
```diff
@@ -83,22 +83,23 @@ Vue.component('global_requirement_general', {
...
-                && !this.details.courses4000Level && !this.details.courses5000Level && !this.details.courses6000Level
-                && !this.details.courses7000Level && !this.details.courses8000Level && !this.details.courses9000Level
-                && this.details.subject_area === "";
-
+            if (is_submission) {
+                this.is_invalid = !this.details.courses1000Level && !this.details.courses2000Level && !this.details.courses3000Level
+                    && !this.details.courses4000Level && !this.details.courses5000Level && !this.details.courses6000Level
+                    && !this.details.courses7000Level && !this.details.courses8000Level && !this.details.courses9000Level
+                    && this.details.subject_area === "";
+            }
```
JackSekoranja: This only changes `is_invalid` when the `is_submission` flag is set

JackSekoranja (changes requested):
Found these issues. Changing the number of units also acts as a submission, so you should make the following change:
`templates/widgets/staff/global_requirements.html:Line 28`:
>`v-on:change="check_options"` to `v-on:change="check_options(false)"`

Also there is no update when the text box is changed, so you should probably do the following change.
`templates/widgets/staff/global_requirements.html:Line 32`:
>`class="text"` to `v-on:change="check_options(false)"`
> (`class="text"` appears twice, hence raising an error)

cassdegrees/static/js/staff/globalrequirements.js
spencerj411: forgot to say in the PR, this is intentional. the reason for this is we dont really want several validation messages popping up at once. The ones that will only pop up when they try to submit are the ones that were instantly popping up.

cassdegrees/static/js/staff/globalrequirements.js
spencerj411: intentional, the message that shows up instatntly depends on is_invalid, hence i added the is_submission condition

JackSekoranja: Still think this needs to be addressed. If they correct their error the error message still appears. This functionality could confuse the user

JackSekoranja (approved):

PR #485, opened 2019-10-13
Remove rules from Course Creation

### LINKED ISSUE #469: Disable Course Rules
In the week 10 team meeting we decided that having rules for courses would be unnecessary for this semesters work, especially as the fields are not updated with Josh's course list search funcitonality. 

We want to disable rules for courses, so they should just have fields for name, code, offerings, etc. (like when creating a course from a program). The existing course rules code should be commented out and left for a future development team. This functionality is more important for when validation becomes a priority.

JackSekoranja:
Closes #469 

Removed advanced rules from courses. Also changed the step count for units in courses to be 6

### DIFF
--- a/cassdegrees/static/js/main.js
+++ b/cassdegrees/static/js/main.js
@@ -140,7 +140,7 @@ document.addEventListener("DOMContentLoaded", () => {
     });
 
     // Capture Vue data when ready
-    if (Vue) {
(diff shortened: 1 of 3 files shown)


Daniel-Jang (approved):
Looks good to me, you can merge it in.

PR #493, opened 2019-11-13
Fix issue with course creation failure

JackSekoranja:
The issue was occurring as the 'rules' field was not present in the form due to deletion of the rules


Daniel-Jang (approved):

JackSekoranja (commented):
Tested and it seems to work.

